PR #1, opened 2022-01-24
Herbivores and Carnivores

MasterpieceElbow:


tests/test_main.py
```diff
@@ -1,6 +1,137 @@
...
+    f = io.StringIO()
+
+    with redirect_stdout(f):
+        print(Animal.alive)
+
+    out = f.getvalue()
+    output = "[{Name: King Lion, Health: 100, Hidden: False}, " \
+             "{Name: Bagira, Health: 100, Hidden: False}, " \
+             "{Name: Susan, Health: 100, Hidden: False}]\n"
+    assert out == output, (
```
IvanRamyk, line 134: JFYI: We decided that error messages are not required in github tasks, so you can leave assertions without them

tests/test_main.py
```diff
@@ -1,6 +1,137 @@
...
+    ], f"'{class_.__name__}' should be inherited from 'Animal'"
+    assert (
+        len(parsed_class.body[0].body) == 1
+    ), f"Only one method '{method}' should be defined inside class '{class_.__name__}'"
+    assert (
+        parsed_class.body[0].body[0].name == method
+    ), f"Only one method '{method}' should be defined inside class '{class_.__name__}'"
+
+
+def test_carnivore_bite_not_hidden():
```
IvanRamyk, line 71: Could you add test case when `Carnivore` tries to bite other `Carnivore`?

IvanRamyk (changes requested):

tests/test_main.py
```diff
@@ -1,6 +1,144 @@
...
+    lion = Animal("Lion King")
+    assert hasattr(lion, "name"), (
+        "Animal instance should have attribute 'name'"
+    )
+    assert hasattr(lion, "health"), (
+        "Animal instance should have attribute 'health'"
+    )
+    assert hasattr(lion, "hidden"), (
+        "Animal instance should have attribute 'hidden'"
+    )
```
danylott, line 27: Code duplication here - you could add attrs = ("name", "health", "hidden") - for better code quality.
For future

IvanRamyk (approved):

danylott (approved):

PR #2, opened 2022-01-25
Solution

Krimat:


app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other):
+        if isinstance(other, Herbivore) and not (other.hidden or self.hidden):
```
phantom-terminal, line 31: You don't need self.hiden here. This makes the bite function a static method.

phantom-terminal (changes requested):

MasterpieceElbow (approved):

PR #6, opened 2022-01-25
Solution

vldtsw:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+class Herbivore(Animal):
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other):
```
alix-mak, line 26: Bad name "other" for the parameter, because you can bite only animal.

alix-mak (changes requested):
Bad name "other" for the parameter, because you can bite only animal.

alix-mak (approved):

app/main.py
vldtsw: Agree with you. Fixed

MasterpieceElbow (approved):

PR #7, opened 2022-01-25
'Solution'

8Men5Camels:


app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name, health=100, hidden=False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
TPolina, line 8: alive is a class attribute so you should reach it from the class and not from the instance of it. same in line 28.

TPolina (changes requested):

app/main.py
8Men5Camels: Doesn't he inherit it from the Animal class? However, the code works as it should. But I will make changes.

TPolina: the code works correctly because when instance doesn't have alive attribute (and it doesn't) it searches for it in class attribute and then changes it. but it's better to change class attribute directly

TPolina (approved):
looks good!

app/main.py
8Men5Camels: I understand. Thanks for the explanation!

MasterpieceElbow (approved):

PR #10, opened 2022-01-25
Solution

Nazar-kob:


Krimat (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, other: Herbivore):
```
liuda-chuzdiuk, line 24: this method should be static and then we could have more fit name than other

liuda-chuzdiuk (changes requested):

liuda-chuzdiuk (approved):

app/main.py
```diff
@@ -1 +1,29 @@
...
+class Herbivore(Animal):
+
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore):
```
MasterpieceElbow, line 25: 'other' isn't good variable here, pick more specific name

MasterpieceElbow (changes requested):

app/main.py
Nazar-kob: Thanks, I changed

MasterpieceElbow (approved):

PR #12, opened 2022-01-25
'Solution'

cloudprogramm:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore):
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
```
Krimat, line 27: Change order of checks so it won`t generate AttributeError with non Herbivore object due to lazy calculation

Krimat (changes requested):

vldtsw (approved):

MasterpieceElbow (approved):

PR #15, opened 2022-01-25
'Solution'

oleksandra-halytska:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name, health=100, hidden=False):
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
```
Nazar-kob, line 8: Sorry, but I think "self.alife" is not good here!

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other):
+        if not other.hidden and not isinstance(other, Carnivore):
+            other.health -= 50
+            if not other.health:
+                other.alive.remove(other)
```
Nazar-kob, line 29: same problems and "if" shows true if "other.health" >= -1

Nazar-kob (changes requested):

Nazar-kob (approved):

MasterpieceElbow (approved):

PR #19, opened 2022-01-25
Solution

nikon4uk:


app/main.py
```diff
@@ -1 +1,26 @@
...
+class Herbivore(Animal):
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal):
+        if isinstance(animal, Herbivore):
+            if not animal.hidden:
```
roland813, line 23: You can use only one loop for and logical operator and

roland813 (approved):

app/main.py
MasterpieceElbow, line 23: You can check isinstance and not hidden in one line

MasterpieceElbow (changes requested):

app/main.py
nikon4uk: thanks!

nikon4uk: thanks!

MasterpieceElbow (approved):

PR #20, opened 2022-01-25
solution

NikitaZH1812:


app/main.py
```diff
@@ -1 +1,30 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self):
```
vldtsw, line 17: You can make it only one line:

vldtsw (changes requested):

app/main.py
NikitaZH1812: thx, fixed

vldtsw (approved):

MasterpieceElbow (approved):

PR #21, opened 2022-01-26
'Solution'

Oleksandr563487953:


app/main.py
```diff
@@ -1 +1,29 @@
...
+        return f"{{Name: {self.name}, Health:" \
+               f" {self.health}, Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self):
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
nikon4uk, line 20: you can do it in one line without any condition

nikon4uk (changes requested):

app/main.py
Oleksandr563487953: Yes, it`s really ugly, thanks. Changes were provided.

app/main.py
```diff
@@ -14,10 +14,7 @@ def __repr__(self):
 
 class Herbivore(Animal):
     def hide(self):
-        if not self.hidden:
-            self.hidden = True
-        else:
-            self.hidden = False
+        self.hidden = False if self.hidden else True
```
nikon4uk, line 17: ok it's batter but you can just do it `self.hidden = not self.hiden` it's same, but without unnecessaty condition.

nikon4uk (approved):

MasterpieceElbow (approved):

PR #23, opened 2022-01-26
Solution

anitaposliedova:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+    def __bool__(self):
+        return self.health > 0
+
+    def __repr__(self):
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+    def dead(self):
```
TPolina, line 20: 'dead' name for function is not very good, because it's hard to understand what function do. check_if_dead is better. or something like this.

app/main.py
```diff
@@ -1 +1,35 @@
...
+        if self in self.alive and self:
```
TPolina, line 21: if i understand your code correctly you remove animal from alive list if its health > 0 but should if health <= 0

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other):
```
TPolina, line 32: 'other' variable name is not very good here, better use 'animal'

TPolina (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+    def __repr__(self):
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+    def check_if_dead(self):
+
+        if self in self.alive and self.health <= 0:
+            self.alive.remove(self)
```
TPolina, line 20: this method should be called from class attribute as well.

TPolina (changes requested):

TPolina (approved):
Well done!

MasterpieceElbow (approved):

PR #29, opened 2022-05-19
Solution

YuriiKindrat:


app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+
+class Herbivore(Animal):
+
+    def hide(self):
+        if self.hidden:
+            self.hidden = False
```
DmytroSvirsa, line 20: it may be done in a more simple way - reverse

DmytroSvirsa (approved):

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def __repr__(self):
+        return "{" + f"Name: {self.name}," \
+                     f" Health: {self.health}," \
+                     f" Hidden: {self.hidden}" + "}"
+
+    def __init__(self, name, health=100, hidden=False):
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
```
Magus6784, line 13: I think better place def __init__ at the start of class. 
We can see class variables (alive and __init__) next to each other, no need to look for what variables in this class - everything at the beginning

app/main.py
Magus6784: Agree

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other):
```
Magus6784, line 28: What does it mean other? Who is other? Other like alien?
I think you got point, other bad name in this case, because we know other is Animal who catched by Carnivore
So we can write like
`victim: Herbivore(or Animal, i don't sure what's better)`

Magus6784 (changes requested):

Magus6784 (approved):

PR #36, opened 2022-05-23
Solution

darten360:


app/main.py
```diff
@@ -1 +1,36 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self):
```
mens0, line 22: you can just write "self.hidden = not self.hidden" and get rid of these 4 lines

mens0 (changes requested):

PR #40, opened 2022-05-23
'Solution'

Yurnerroo:


app/main.py
```diff
@@ -1 +1,27 @@
...
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal):
+        if isinstance(animal, Herbivore) and animal.hidden is False:
+            animal.health -= 50
+            if not animal.health:
```
MasterpieceElbow, line 26: What if Herbivore's health will become `-10` after the bite?

MasterpieceElbow (changes requested):

PR #42, opened 2022-05-23
All done.

Flashmobber:


app/main.py
```diff
@@ -1 +1,29 @@
...
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(target: Herbivore):
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
```
darten360, line 27: it's not nescessary to use loop in loop here. Try to use them separately

app/main.py
```diff
@@ -1 +1,29 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        return "{" + f"Name: {self.name}, " \
```
darten360, line 11: You could use f string with curvy brace inside, it's good practice

darten360 (changes requested):

app/main.py
Flashmobber: You've said loop? But I hope you did meant if?

darten360 (approved):

PR #46, opened 2022-05-23
Herbivores and carnivores solution

Ed12816:
even in OOP you can become a target for a carnivores  |O_O|

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


Y-Havryliv (approved):

app/main.py
```diff
@@ -1 +1,28 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(target: Herbivore):
+        if not target.hidden and isinstance(target, Herbivore):
+            target.health -= 50
+            if target.health == 0:
```
MasterpieceElbow, line 27: What if after the bite Herbivore health become `-10`?

MasterpieceElbow (changes requested):

app/main.py
Ed12816: let health be a negative value if necessary

MasterpieceElbow (approved):

PR #54, opened 2022-05-23
Solution

NOSSUDAK:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
lenasharabura, line 8: I think it's better to check if the animal is alive before adding it to the list of the alive.

lenasharabura (approved):

MasterpieceElbow (approved):

PR #55, opened 2022-05-23
'Solution'

Aleksei-Isaev:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health=100, hidden=False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
```
lenasharabura, line 9: I think it's better to check if the animal is alive before adding it to the list of the alive.

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal):
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health -= 50
+            if not animal.health:
```
lenasharabura, line 28: not very good conditions because i can create an animal with 30 xp, after biting it still will be in alive list in this case

lenasharabura (changes requested):
Good job, but i found some mistaces.

MasterpieceElbow (approved):

PR #61, opened 2022-05-23
solution

Y-Havryliv:


NOSSUDAK (approved):
In line 29 you can use <= instead of == because it would be more reliable. Imagine someone will create animal with health = 99 :)

mens0 (approved):

app/main.py
```diff
@@ -1 +1,27 @@
...
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other):
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+        if other.health == 0:
```
MasterpieceElbow, line 26: Imagine Carnivore bites Herbivore with 25 health points, will that Herbivore die?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #63, opened 2022-05-23
Solution

Chardona:


app/main.py
```diff
@@ -1 +1,32 @@
...
+class Herbivore(Animal):
+
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore):
```
Kiawami, line 28: it was mentioned that 'other' isn't good variable here (you can bite only animal)

Kiawami (changes requested):

app/main.py
Chardona: that's why my next line is a check up for  the correct animal.

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            if other.health == 0:
```
katevasilenko, line 31: I made this mistace too) What if bitten animal would have health below zero

katevasilenko (changes requested):

app/main.py
Chardona: We make hp = 100 by default. How can we get hp < 0, if we only make steps with scale of 50? But it's truly better to take care of unpredicted case.

Chardona: Thanks, that's a  good point. As it is said "If person can make a mistake, he'll definitely do it"

MasterpieceElbow (approved):

PR #65, opened 2022-05-23
Solution

Evgeniy994:


app/main.py
```diff
@@ -1 +1,28 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore):
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
+            herbivore.health -= 50
+        if herbivore.health == 0:
```
rotsen18, line 27: What if Herbivore's health will become -10 after the bite?

rotsen18 (changes requested):

PR #66, opened 2022-05-23
Solution

Kiawami:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal):
+        if isinstance(animal, Herbivore) and animal.hidden is False:
+            animal.health -= 50
+            if animal.health == 0:
```
rotsen18, line 30: What if Herbivore's health will become -10 after the bite?

rotsen18 (changes requested):

PR #68, opened 2022-05-24
Develop

vasylhnatiuk:


app/main.py
```diff
@@ -1 +1,31 @@
...
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore):
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
+            herbivore.health = herbivore.health - 50
```
katevasilenko, line 29: you can use -= to avoid repeating

app/main.py
```diff
@@ -1 +1,31 @@
...
+        if herbivore.health == 0:
```
katevasilenko, line 30: I made this mistake too) what if bitten animal would have health below zero?

katevasilenko (changes requested):

app/main.py
vasylhnatiuk: thanks)

katevasilenko (approved):

MasterpieceElbow (approved):

PR #84, opened 2022-06-17
Solution

muhactive:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __repr__(self):
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self):
+        if self.hidden is False:
```
IvanRamyk, line 19: You can simplify this logic:
```
self.hidden = not self.hidden
```
But it's also OK

IvanRamyk (approved):

PR #88, opened 2022-06-29
Solution

VladyslavKlevanskyi:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        return "{Name: " + self.name + \
```
Abnormaltype, line 12: please, use interpolation

app/main.py
```diff
@@ -1 +1,34 @@
...
+               ", Health: " + str(self.health) + \
+               ", Hidden: " + str(self.hidden) + "}"
+
+
+class Herbivore(Animal):
+
+    def hide(self):
+        if not self.hidden:
+            self.hidden = True
+        else:
```
Abnormaltype, line 22: you can write this without `else`

Abnormaltype (changes requested):

app/main.py
VladyslavKlevanskyi: I tried but don't know how to exclude the curly brace at the beginning and end of the string. The backslash doesn't work.
изображение

Abnormaltype: `        return f"{{Name: {self.name}," \
               f" Health: {self.health}," \
               f" Hidden: {self.hidden}}}"

`
try this variant

app/main.py
VladyslavKlevanskyi: I'm not sure how. If you mean this construction, then it will not work. Because the new True value will be overwritten by False
изображение

Abnormaltype (changes requested):
suggestion

app/main.py
VladyslavKlevanskyi: Thank you, but it is seems to me not easier than what I did :)

app/main.py
VladyslavKlevanskyi: Cunningly ))

app/main.py
Abnormaltype: but looks better

Abnormaltype (approved):
good job!

PR #97, opened 2022-07-11
Solution

InnaKushnir:


app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def __init__(self, name, health=100, hidden=False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        a = f"Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}"
+        return f'{{{a}}}'
```
Abnormaltype, line 13: what is `a`?

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self):
+        if self.hidden is False:
+            self.hidden = True
+            return self.hidden
+        self.hidden = False
+        return self.hidden
```
Abnormaltype, line 23: you can write this in one line

Abnormaltype (changes requested):

app/main.py
InnaKushnir: return self.hidden is True if self.hidden is False else self.hidden is False
Does not pass the tests when I use the ternary operator

Abnormaltype: try to write like this `self.hidden = not self.hidden`

app/main.py
InnaKushnir: Removed this crutch

InnaKushnir: Line break error 12, cannot break a line correctly

Abnormaltype (changes requested):
fix all mistakes

Abnormaltype (changes requested):

Abnormaltype (approved):
well done

PR #98, opened 2022-07-12
Solution

DHushchyk:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        if other.health <= 0:
+            Animal.alive.remove(other)
+
+    def __repr__(self):
+        name = f"Name: {self.name}"
+        health = f"Health: {self.health}"
+        hidden = f"Hidden: {self.hidden}"
+        f1 = "{"
+        f2 = "}"
+        return f"{f1}{name}, {health}, {hidden}{f2}"
```
Abnormaltype, line 22: write this, without variables

Abnormaltype (changes requested):

app/main.py
DHushchyk: It is because of "A single '{'  or '}' is not allowed inside f-strings"

DHushchyk: image
image

Abnormaltype: image

PR #101, opened 2022-07-18
Solution

Povshok:


app/main.py
```diff
@@ -1 +1,53 @@
...
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+        if target.health <= 0:
+            Animal.alive.remove(target)
+
+
+class Herbivore(Animal):
+
+    def hide(self):
+        # if self.hidden is True:
```
IvanRamyk, line 29: Comments shouldn't be left in PR

app/main.py
```diff
@@ -1 +1,53 @@
...
+        #     self.hidden = False
+        # else:
+        #     self.hidden = True
+        self.hidden = not self.hidden
+
+# pantera = Carnivore("Bagira")
```
IvanRamyk, line 35: These tests shouldn't be left too

IvanRamyk (changes requested):

app/main.py
Povshok: is everything ok?

IvanRamyk (approved):
Well done!

PR #108, opened 2022-08-05
Solution

anastasia-martyniuk:


app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self):
+        if self.hidden is False:
+            self.hidden = not False
+        elif self.hidden is True:
+            self.hidden = not True
+        return self
```
Abnormaltype, line 24: you can write this in one line

Abnormaltype (changes requested):

app/main.py
anastasia-martyniuk: I changed this func to this:
def hide(self):
        self.hidden = True if self.hidden is False else False
        
and now I have a proble with pytest, but I don't understand why?(

Abnormaltype: Try to write like this
` self.hidden = not  self.hidden`

anastasia-martyniuk: image
why am I getting these?

Abnormaltype: push this code please

Abnormaltype (changes requested):
remove all changes in README and tests

Abnormaltype (approved):
Well done!

PR #115, opened 2022-08-17
Solution

KatyaVasylieva:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self, name, health=100, hidden=False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
```
DmytroSvirsa, line 9: to provide less mistakes better use `Animal` instead of `self.__class__`

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, prey):
+        if not isinstance(prey, Carnivore) and prey.hidden is False:
+            prey.health -= 50
+            if prey.health <= 0:
+                self.alive.remove(prey)
```
DmytroSvirsa, line 29: to provide less mistakes better use `Animal` instead of `self.__class__`

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #118, opened 2022-08-22
Solution

Paul-Starodub:


app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False):
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
+
+    def __repr__(self):
+        return f"{'{'}Name: {self.name}, "\
+            f"Health: {self.health}, Hidden: {self.hidden}{'}'}"
```
Abnormaltype, line 12: would be better to write like this:
```suggestion
        return f"{{Name: {self.name}," \
               f" Health: {self.health}," \
               f" Hidden: {self.hidden}}}"
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #130, opened 2022-09-08
'Solution'

yevhenii-nevmyvako:


app/main.py
```diff
@@ -1 +1,29 @@
...
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self):
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
DmytroSvirsa, line 21: You can change the `bool` value in one line:

Good example:

```python
is_married = True
is_married = not is_married
```

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, other):
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health = other.health - 50
+            if other.health < 1:
```
DmytroSvirsa, line 28: try avoid nested constructions

DmytroSvirsa (changes requested):

app/main.py
yevhenii-nevmyvako: i cannot chenge, I already tried initially, but with this solution it gives an error. (AssertionError: Method 'hide' should change animal attribute 'hidden')
class Herbivore(Animal):
    def hide(self):
        if self.hidden is False:
            self.hidden = not self.hidden

yevhenii-nevmyvako: image

app/main.py
yevhenii-nevmyvako: revmove the nested construction

app/main.py
yevhenii-nevmyvako: understood, I changed

Abnormaltype (approved):
Well done!

PR #133, opened 2022-09-11
we have class Animal which create animal and put to alive list. Class…

LitvinOleksandr:
… carnivore extended from animal, and has function bite. Carnivore can't bite carnivore. Every time after bite health - 50 points. Class herbivore extendet from animal class, and has func hide. If animal has hide status True carnivore can't bite. We have mesage about this.

If some animal has health less than 0 automatically deleted from this list

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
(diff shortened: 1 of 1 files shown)


Abnormaltype (approved):
Good job! If you wanna describe your code, write a docstring)

PR #145, opened 2022-09-18
Solution

mltdll:


app/main.py
```diff
@@ -1 +1,45 @@
...
+            self.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
```
AVyha, line 42: I think you need use @staticmethod, you don't use `self` in this function.

AVyha (changes requested):

app/main.py
mltdll: I don't think so: static method may be called directly from the class.
`Carnivore.bite()`, however, is intended to be used only from class instances.

AVyha: You still can call this function with class instances when this function have @staticmethod, and with this method you don't have unuse `self`. I think it's best practise.

AVyha: Now you can call its function with class instances(for example `lion.bite(rabit)`) but can't with class `Carnivore.bite(rabit)`.
With @staticmethod you can call with class instances and with class.

AVyha (approved):
I still think that @staticmethod is best practise but maybe im wrong)) Maybe mentor say who right in this question.

app/main.py
mltdll: > You still can call this function with class instances when this function have @staticmethod

Yes, I know. But the reason I think the method shouldn't be static is that I don't want it to be used directly from the class. E.g.:
```python
Carnivore.bite(rabbit)  # TypeError: Carnivore.bite() missing 1 required positional argument: 'other'
```

AVyha: Ok, maybe you right)  Sorry for that remark

Y-Havryliv (approved):
Well done!

PR #146, opened 2022-09-18
Solution

KirillDrago:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        Animal.alive.append(self)
+
+    def not_alive(self):
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self):
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
sbublinskyi, line 17: I recommend you to use ( ) for multiline strings instead of backslash \

sbublinskyi (changes requested):

sbublinskyi (approved):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal):
```
TPolina, line 28: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name, health=100, hidden=False):
```
TPolina, line 4: add type annotation everywhere

TPolina (changes requested):

TPolina (approved):
Well done!

PR #149, opened 2022-09-20
Solution

kateryna3k:


AriohBelskij (approved):
Good!

app/main.py
```diff
@@ -1 +1,29 @@
...
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
TPolina, line 25: you don't need `self` here, so make this method static

TPolina (changes requested):

TPolina (approved):

PR #153, opened 2022-09-20
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self):
```
wQuelS, line 18: add type annotation to function

app/main.py
```diff
@@ -1 +1,28 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal) -> None:
+        if isinstance(animal, Herbivore) and animal.hidden is False:
+            animal.health -= 50
+            if animal.health <= 0:
+                del self.alive[self.alive.index(animal)]
```
wQuelS, line 28: consider to use remove(). Index may be irreliable sometimes. But it is up to you

wQuelS (changes requested):

app/main.py
taras-andruschenko: I prefer to use "del", but thx for your suggestion

wQuelS (approved):
Let's hope mentros will not reject due to index usage in the last line)

taras-andruschenko:
Will see :)

app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health=100, hidden=False) -> None:
```
TPolina, line 4: add type annotation everywhere

app/main.py
```diff
@@ -1 +1,28 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal) -> None:
+        if isinstance(animal, Herbivore) and animal.hidden is False:
+            animal.health -= 50
+            if animal.health <= 0:
+                del self.alive[self.alive.index(animal)]
```
TPolina, line 28: simplify this line. We access class attributes through class name and it's better to use `remove` when value is given and `del` when an index is given. You do unnecessary work here and readability is bad

TPolina (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Animal) -> None:
```
Nattalli, line 26: it have to be a `@staticmethod`

Nattalli (changes requested):

app/main.py
taras-andruschenko: agree

PR #156, opened 2022-09-21
Solution

Aleksandro777:


app/main.py
```diff
@@ -1 +1,33 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is True:
+            self.hidden = False
+        else:
+            self.hidden = True
```
sbublinskyi, line 22: you can change bool type, like this:
self.hidden = not self.hidden

sbublinskyi (changes requested):
everything else is good

TPolina (approved):
Good job!

PR #159, opened 2022-09-21
Solution)

Gr1zzZzzly:


mltdll (approved):

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore):
+        herbivore.health -= (50 if isinstance(herbivore, Herbivore) and
+                             herbivore.hidden is False else 0)
```
TPolina, line 34: Why do you use the ternary operator here? Simple if block will be simpler and more readable

TPolina (changes requested):

TPolina (approved):
Well done!

PR #165, opened 2022-09-22
solve the task

Cerne13:


app/main.py
```diff
@@ -1 +1,29 @@
...
+class Animal:
+
+    alive = []
+
+    def __repr__(self) -> str:
+        return f'{{Name: {self.name}, ' \
+               f'Health: {self.health}, ' \
+               f'Hidden: {self.hidden}}}'
+
+    def __init__(self, name, health=100, hidden=False):
```
Kitazuka, line 10: Add name annotation

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __repr__(self) -> str:
+        return f'{{Name: {self.name}, ' \
+               f'Health: {self.health}, ' \
+               f'Hidden: {self.hidden}}}'
```
Kitazuka, line 8: You can use () instead of /, but I'm not sure if this is best practice

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    def __init__(self, name, health=100, hidden=False):
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+
+class Herbivore(Animal):
+    def hide(self):
```
Kitazuka, line 18: and add an annotation for the return "hide"

Kitazuka (changes requested):

app/main.py
Cerne13: Thank you!

app/main.py
Cerne13: I'll try to leave it as is for now (imho, more readable in scope of curly brackets symbol escaping)

app/main.py
```diff
@@ -1 +1,29 @@
...
+    def __repr__(self) -> str:
+        return f'{{Name: {self.name}, ' \
+               f'Health: {self.health}, ' \
+               f'Hidden: {self.hidden}}}'
+
+    def __init__(self, name: str, health=100, hidden=False):
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
```
TPolina, line 14: init should be the first method of a class

TPolina (changes requested):

app/main.py
Cerne13: Thank you. I somehow overlooked this.

TPolina (approved):

PR #171, opened 2022-09-22
Solution

skrypka-dmytro:


app/main.py
```diff
@@ -1 +1,33 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def death(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
```
ketstap162, line 20: Delete func `death`

app/main.py
```diff
@@ -1 +1,33 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health -= 50
+            animal.death()
```
ketstap162, line 33: Add body of func `death` here

ketstap162 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
ketstap162, line 16: Don't use `\`
Create a temporary variables

ketstap162 (approved):

ketstap162 (approved):
Well done!

app/main.py
Y-Havryliv: it' better to use `()` in multiline strings

app/main.py
Y-Havryliv: it' better to use `()` in multiline strings

Y-Havryliv (approved):
GJ!

PR #173, opened 2022-09-23
'Solution'

OvDima1:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False):
```
serviza, line 8: You can add annotation to this method.

serviza: I have no questions more)

serviza (changes requested):

serviza (approved):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not herbivore.hidden and isinstance(herbivore, Herbivore):
```
Y-Havryliv, line 28: it's better to check on `isinstance` first

Y-Havryliv (changes requested):

Nattalli (approved):
👍🏻

PR #175, opened 2022-09-23
Solution

VitaliiPiaskivskyi:


app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> None:
```
Vadosh, line 6: This line breaks doesn't improve the readability of the code in my opinion. Will look better in one line, or every attribute on new line.

Vadosh (approved):

app/main.py
```diff
@@ -1 +1,38 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
```
Nattalli, line 15: do you really need these variables? please remove them as you can do the same without them

app/main.py
```diff
@@ -1 +1,38 @@
...
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        for number, one_anim in enumerate(Animal.alive):
+            if one_anim.name == self.name and isinstance(one_anim, Herbivore):
+                one_anim.hidden = False if one_anim.hidden else True
+                Animal.alive[number] = one_anim
```
Nattalli, line 25: you can make it more readable and simply:
```python
self.hidden = not self.hidden
```
- it makes the same as you made

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def bite(self, herbivore: Animal) -> None:
+        for number, one_a in enumerate(Animal.alive):
+            if one_a.name == herbivore.name and isinstance(one_a, Herbivore):
+                if not one_a.hidden:
+                    one_a.health -= 50
+                    if one_a.health == 50:
+                        Animal.alive[number] = one_a
+                    else:
+                        Animal.alive.pop(number)
```
Nattalli, line 38: here you can make better code too
you have to make the first `if` condition which checks if the `herbivore` is instance of `Herbivore` and if `herbivore.hidden` is equal to False. And then make `herbivore.health -= 50`

The second if checks if `herbivore.health <= 0` and then `remove` it form the `Animal.alive` list

Nattalli (changes requested):

Nattalli (approved):
Awesome! 😎

PR #179, opened 2022-09-23
Solution

linaSam:


YevheniiDavydenko (approved):
I like it
hope mentor approve too

app/main.py
```diff
@@ -1 +1,32 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
```
Y-Havryliv, line 19: you can change bool value using `bool = not bool`

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
Nattalli, line 23: Yaroslav told that all these 4 lines are equal to one line:
```python 
bool = not bool
```
So you can replace it

Nattalli (changes requested):

PR #184, opened 2022-09-23
Solution

pavlo-kapitanets:


app/main.py
```diff
@@ -1 +1,29 @@
...
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
```
FranchukS, line 25: You don`t need self in this method, so you can use staticmethod

FranchukS (changes requested):

app/main.py
pavlo-kapitanets: Done, thanks for your review.

FranchukS (approved):
GJ!

app/main.py
```diff
@@ -1 +1,30 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if animal.hidden is False and isinstance(animal, Herbivore):
+            animal.health -= 50
+            if animal.health <= 0:
```
Nattalli, line 29: this condition has to be at the same level with the first condition

Nattalli (changes requested):

Nattalli (approved):
Good job! 👍🏻

Nattalli (approved):

PR #187, opened 2022-09-24
Solution

fsocie7y:


app/main.py
```diff
@@ -1 +1,28 @@
...
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other) -> None:
```
matechris, line 23: I think this is a static method

matechris (changes requested):

matechris (approved):
Good job :)

app/main.py
```diff
@@ -1 +1,28 @@
...
+    def __init__(
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name, health, hidden = self.name, self.health, self.hidden
```
Nattalli, line 13: does it needed variables? You can make the same options without them

Nattalli (changes requested):

app/main.py
fsocie7y: whith out them flake8 will be crash

fsocie7y: ok, i`m fixed it

Nattalli (approved):
Cool!

PR #191, opened 2022-09-25
'Solution'

Menselface:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health=100, hidden=False) -> None:
```
linaSam, line 4: i would put spaces near =

linaSam (changes requested):

linaSam (approved):
Great!

app/main.py
```diff
@@ -1 +1,34 @@
...
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{" \
+               f"Name: {self.name}, " \
```
Nattalli, line 17: not readable, change for:
```python
return (f"{{Name: {self.name}, ....")
```

Nattalli (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{"f"Name: {self.name}, "
```
Nattalli, line 16: and one more: don't use 2 f-strings, one is enough

Nattalli (changes requested):

Nattalli (approved):
Wow! Now it's good

PR #194, opened 2022-09-25
'Solution'

illiariaboy94:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self):
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Animal):
```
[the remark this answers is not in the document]
illiariaboy94: Ok. Thanks for review

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False)\
+            -> None:
```
rostyak, line 5: don't use "\", better do something like this:
    def __init__(self, name: str, health: int = 100, 

rostyak (approved):

app/main.py
illiariaboy94: Done.

rostyak: Good job

Nattalli (approved):
👍🏻

PR #198, opened 2022-09-26
Solution

mariia-movchan:


rostyak (approved):

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Carnivore)\
+                and not herbivore.hidden:
+            herbivore.health -= 50
+
+        if herbivore.health < 1:
```
MasterpieceElbow, line 35: It is more clear and explicit to write `<=0` rather than `<1`. Because animal with 0.5 health is still alive.

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Carnivore)\
+                and not herbivore.hidden:
```
MasterpieceElbow, line 32: Avoid using `\`.
```suggestion
if (not isinstance(herbivore, Carnivore)
     and not herbivore.hidden):
```

MasterpieceElbow (changes requested):

app/main.py
mariia-movchan: fixed

app/main.py
mariia-movchan: Suggested solution leads to W503 line break before binary operator. But I fixed it another way

TPolina (approved):
Well done!

PR #200, opened 2022-09-29
solution

marina-bidenko:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+    def __init__(self, name: str, health: int = 100):
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self):
+        return "{" + f"Name: {self.name}, \
+Health: {self.health}, \
```
MasterpieceElbow, line 13: Avoid using `\`. Why do you use string concatenation if you can just write it all within one f-string?

PR #205, opened 2022-10-07
Herbivores and carnivores

LavrykN:


app/main.py
```diff
@@ -1 +1,36 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Animal) -> None:
+        if isinstance(herbivore, Herbivore):
+            if not herbivore.hidden:
+                herbivore.health -= 50
+            self.viability()
```
IvanRamyk, line 36: Why check all `animals` for a single bite? Check only animal that was bit

IvanRamyk (changes requested):

app/main.py
LavrykN: Agree, corrected

IvanRamyk (approved):
Good job!

PR #206, opened 2022-10-08
Solution of py-herbivores-and-carnivores

YLobunets:


app/animals.py
```diff
@@ -0,0 +1,42 @@
...
+        return "{" + f"Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}" + "}"
+
+    @staticmethod
+    def add_born(animal: Animal) -> None:
+        Animal.alive.append(animal)
+
+    @staticmethod
+    def remove_dead(animal: Animal) -> None:
+        for index in range(len(Animal.alive)):
```
IvanRamyk, line 25: use `remove` method instead

IvanRamyk (changes requested):

app/animals.py
YLobunets: Method replaced

app/animals.py
```diff
@@ -0,0 +1,42 @@
...
+            if animal is Animal.alive[index]:
+                Animal.alive.remove(animal)
```
IvanRamyk, line 27:
```suggestion
      Animal.alive.remove(animal)

```

IvanRamyk: You can call this method without using a cycle

IvanRamyk (changes requested):

app/animals.py
YLobunets: Yes, you are right, I changed the function, but the code was not optimized.
Cycle removed

IvanRamyk (approved):
wd

PR #215, opened 2022-10-16
Solution

Mazurinka29:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def __repr__(self) -> str:
+        alive_animal = f"{{Name: {self.name}, Health: {self.health}," \
+                       f" Hidden: {self.hidden}}}"
+        return alive_animal
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = self.hidden
```
DmytroSvirsa, line 20: unnessasary line

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: type) -> None:
+        if herbivore.hidden or isinstance(herbivore, Carnivore):
+            herbivore.health = herbivore.health
+        else:
+            herbivore.health -= 50
```
DmytroSvirsa, line 31: use not to avoid else statement

DmytroSvirsa (changes requested):

app/main.py
DmytroSvirsa: i mean that self.hidden = self.hidden is not needed

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):
well done!

PR #221, opened 2022-10-26
Develop

Soobig666:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        animals = Animal.alive
+        if not isinstance(animal, Carnivore) and not animal.hidden:
+            animal.health -= 50
+            if animal.health <= 0:
+                animals.pop(animals.index(animal))
```
Serhii-Leonenko, line 35: Do you remember about `list.remove()`?)

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
Abnormaltype, line 25: you can write this in one line

Abnormaltype (changes requested):

app/main.py
Soobig666: yes, but i like use list.pop(), cos sometimes i need to use this data was i delete.
But, i will fix it for you

app/main.py
Soobig666: kk, boss. I will do it

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = True if not self.hidden else False
```
Abnormaltype, line 22:
```suggestion
        self.hidden = not self.hidden
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #244, opened 2022-11-21
Solution

T-Vlad-96:


app/main.py
```diff
@@ -1 +1,37 @@
...
+            self.health = args[0]
+            self.hidden = args[1]
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{"\
+               + f"Name: {self.name}," \
+                 f" Health: {self.health}," \
+                 f" Hidden: {self.hidden}"\
+               + "}"
```
Abnormaltype, line 23: use interpolation

app/main.py
```diff
@@ -1 +1,37 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    health = 100
+    hidden = False
```
Abnormaltype, line 6: it should be attributed to an instance

Abnormaltype (changes requested):

app/main.py
Abnormaltype: what about it?

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def __init__(self, name: str, *args, **kwargs) -> None:
+        self.name = name
+        if len(args) == 1:
+            self.health = args[0]
+        elif self.health == 2:
+            self.health = args[0]
+            self.hidden = args[1]
+        else:
+            self.health = 100
+            self.hidden = False
```
Abnormaltype, line 18: remove all if/else

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #249, opened 2022-11-25
Solution

andriy-md:


app/main.py
```diff
@@ -1 +1,38 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if not isinstance(victim, Herbivore):
+            return
+        if victim.hidden:
+            return
+        victim.health -= 50
+        victim.check_health()
```
YuriiKindrat, line 38: No need to use blank returns here, u can make only one check for case, when u should make damage for victim

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #252, opened 2022-11-28
solution

anastasiia-tsurkan:


app/main.py
```diff
@@ -1 +1,38 @@
...
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+
+    def bite(self, other: Herbivore) -> None:
+        if isinstance(other, Herbivore):
+            if not other.hidden:
```
YuriiKindrat, line 35: U can use only one if here

app/main.py
```diff
@@ -1 +1,38 @@
...
+            f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
YuriiKindrat, line 28: U can use here more simple construction, like `self.value = not self.value`

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #257, opened 2022-12-02
Solution

Nexxel05:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Carnivore) or animal.hidden is True:
+            pass
+        else:
+            animal.health -= 50
+            if animal.health <= 0:
+                Animal.alive.remove(animal)
```
YuriiKindrat, line 39: No need to use else here, u can write other condition for "if"

app/main.py
```diff
@@ -1 +1,39 @@
...
+        return Animal(self.name, self.health, self.hidden)
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
+        return self.hidden
```
YuriiKindrat, line 28: No need to use return here, the function just change the status of hidder attribute, also u can write it in one line like:
self.hidden = not self.hidden

app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
YuriiKindrat, line 8: fix the breaking line here according to the best practices

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        if self.hidden is self.hidden:
+            self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Carnivore) or animal.hidden is True:
+            pass
```
YuriiKindrat, line 33: No need of this if

app/main.py
```diff
@@ -1 +1,37 @@
...
+        if isinstance(animal, Herbivore) and animal.hidden is False:
```
YuriiKindrat, line 34: u can write "not animal.hidden" instead of "is False"

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+    def __sub__(self, other: Animal) -> Animal:
+        self.health -= other.health
+        return Animal(self.name, self.health, self.hidden)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is self.hidden:
+            self.hidden = not self.hidden
```
YuriiKindrat, line 26: No need of this if, u can just write "self.hidden = not self.hidden"

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, Hidden: {False}}}"
+
+    def __sub__(self, other: Animal) -> Animal:
+        self.health -= other.health
+        return Animal(self.name, self.health, self.hidden)
```
Serhii-Leonenko, line 20: It's an unnecessary method here

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
andrewbelo, line 13: краще використати тут self.__class__

andrewbelo (changes requested):

app/main.py
andrewbelo: з подвійними андерскорами (__ __)

IvanRamyk (approved):

PR #259, opened 2022-12-07
done

phaishuk:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        for beast in Animal.alive:
+            if beast.name == self.name:
```
Abnormaltype, line 25: unnecessary

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
OleksandrPozden, line 10:
```suggestion
    def __init__(
               self,
                 name: str,
                 health: int = 100,
                 hidden: bool = False 
       ) -> None:
```

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(some_animal: Herbivore) -> None:
+        if isinstance(some_animal, Herbivore) and \
+                some_animal.hidden is False:
```
OleksandrPozden, line 32:
```suggestion
        if isinstance(some_animal, Herbivore) and \
                not some_animal.hidden:
```
Also, use brackets, instead of ```\```

OleksandrPozden (changes requested):

app/main.py
phaishuk: That's better because of code style?

Abnormaltype (approved):

PR #264, opened 2022-12-10
Solution

Isles22:


app/main.py
```diff
@@ -1 +1,44 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
```
OleksandrPozden, line 31:
```suggestion
       self.hidden = not self.hidden
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def bite(self, beast: Animal) -> None:
+        if isinstance(beast, Herbivore) and beast.hidden is False:
+            if beast.health > 50:
+                beast.health -= 50
+            else:
+                beast.health = 0
+
+                for i in range(len(beast.alive)):
+                    if beast.alive[i] == beast:
+                        beast.alive.pop(i)
```
OleksandrPozden, line 44:
```suggestion
            beast.health -= 50
            if beast.health<0:
                 Animal.alive.remove(beast)
```
can we have that?

OleksandrPozden (changes requested):

app/main.py
Isles22: of course it's better

app/main.py
Isles22: I agree

IvanRamyk (approved):
Well done!

PR #266, opened 2022-12-10
Solution

SlavikSherbak:


app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, *args) -> None:
+        self.name = name
+        self.health = 100 if args == () else args[0]
+        self.hidden = False
+        Animal.alive.append(self)
```
OleksandrPozden, line 8:
```suggestion
    def __init__(self, name: str, health, hidden=False) -> None:
        self.name = name
        self.health = health
        self.hidden = hidden
        Animal.alive.append(self)
```

OleksandrPozden: but dont forget about annotation

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Vasyl-Smutok, line 18: better to use `()` then `/`
```suggestion
        return (f"{{Name: {self.name}, " 
               f"Health: {self.health}, " 
               f"Hidden: {self.hidden}}}")
```

Vasyl-Smutok (changes requested):

katevasilenko (approved):
Good job!

PR #278, opened 2022-12-28
Solution

olenaliuby:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{" \
+               f"Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}" \
```
katevasilenko, line 18: Please, avoid using `\` if it possible

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is True:
+            self.hidden = False
+        else:
+            self.hidden = True
```
katevasilenko, line 31: You can write it in one line. The hint is: use `not`

katevasilenko (changes requested):
Looks good just need a bit improve

app/main.py
olenaliuby: Thank you! Changed to just one `\`, can't delete all cause of flake8.

app/main.py
olenaliuby: Thank you! Totally forget about this case from checklist. :(

IvanRamyk (approved):

PR #284, opened 2023-01-02
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{" + f"Name: {self.name}, Health: {self.health}," \
+                     f" Hidden: {self.hidden}" + "}"
```
Abnormaltype, line 12: use interpolation

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
Abnormaltype, line 20: you can write this in one line

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
```
katevasilenko, line 17: `self.hidden =  not self.hidden`
Does that make sense?

app/main.py
```diff
@@ -1 +1,27 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
```
katevasilenko, line 11: Please avoid using `\` whenever it possible

app/main.py
```diff
@@ -1 +1,27 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other_animal: Animal) -> None:
+        if isinstance(other_animal, Herbivore) and \
+                other_animal.hidden is False:
```
katevasilenko, line 24: `and` Should be at a new 24 line an avoid \

katevasilenko (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
```
Vasyl-Smutok, line 14: no need to create these variables
you can do it like this
`return f"{{Name: {self.name}..."`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
Vasyl-Smutok, line 7: I think you should add `hidden` in `__init__`

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+class Animal:
+    alive = []
```
ArsenPidhoretskyi, line 2: Add type hint. 
HINT: use from __future__ import annotations

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: " \
+               f"{self.health}, Hidden: {self.hidden}}}"
```
Vasyl-Smutok, line 15: better like this
``` 
        return (f"{{Name: {self.name}, Health: " 
                f"{self.health}, Hidden: {self.hidden}}}")
```

app/main.py
Vasyl-Smutok: please do as @ArsenPidhoretskyi said

here is an example for you
`alive: list[Animal] = [] `

Vasyl-Smutok (changes requested):

katevasilenko (approved):
Looks good to me

PR #294, opened 2023-01-12
'Solution'

StinGER-ua:


app/main.py
```diff
@@ -1 +1,33 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
+    def bite(self, animal_other: Herbivore) -> None:
```
olena-hul, line 28: And name it `other_animal`, sounds more natural, or `herbivore`

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
```
olena-hul, line 18:
```suggestion
        return (
            f"{{Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}}}"
        )
```

olena-hul (changes requested):

PR #297, opened 2023-01-13
Solution

PodorogaNatalia:


app/main.py
```diff
@@ -1 +1,36 @@
...
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
```
Vasyl-Smutok, line 10: in my opinion, it is better to write it like this
`Animal.alive.append(self)`

everything else is written very cool)

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
```
Vasyl-Smutok, line 28: also here you can write it like this

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #299, opened 2023-01-15
py-herbivores-and-carnivores solution

frameq:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return str(
+            {"Name": self.name, "Health": self.health, "Hidden": self.hidden}
+        ).replace("'", "")
```
olena-hul, line 19: Why do you need replace? Remove it

olena-hul (changes requested):

app/main.py
frameq: without replace tests will fail:
`E       assert "[{'Name': 'K...n': False}]\n" == '[{Name: King...en: False}]\n'
E         - [{Name: King Lion, Health: 100, Hidden: False}, {Name: Bagira, Health: 100, Hidden: False}, {Name: Susan, Health: 100, Hidden: False}]
E         + [{'Name': 'King Lion', 'Health': 100, 'Hidden': False}, {'Name': 'Bagira', 'Health': 100, 'Hidden': False}, {'Name': 'Susan', 'Health': 100, 'Hidden': False}]
E         ?   +    +  +         +  +      +       +      +           +    +  +      +  +      +       +      +           +    +  +     +  +      +       +      +
`

olena-hul: Anyway, the solution could be done in a cleaner way. Use f-strings instead of str constructor and .replace()

olena-hul (changes requested):

Abnormaltype (approved):
Well done!

PR #301, opened 2023-01-15
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1 +1,30 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
+        return f"""{{Name: {name}, Health: {health}, Hidden: {hidden}}}"""
```
olena-hul, line 14: No need to put those in variables. We can access it directly from self

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
olena-hul, line 22:
```suggestion
        self.hidden = not self.hidden
```

olena-hul (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #302, opened 2023-01-16
Solution

Andrii-Kozlovskyi:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.hidden = False
```
Vasyl-Smutok, line 6: think it should be added to `__init__`

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.health = health
+        self.__class__.alive.append(self)
```
Vasyl-Smutok, line 8: use    `Animal.alive.append(self)`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
```
ArsenPidhoretskyi, line 2: Specify the type for the attribute. 
HINT: You can use `from __future__ import annotation`

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #303, opened 2023-01-17
Solution

avkpol:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
```
ArsenPidhoretskyi, line 2: Please specify type of property

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Herbivore):
+            if animal.hidden is False:
+                animal.health = animal.health - 50
+                if animal.health <= 0:
+                    Animal.alive.remove(animal)
```
ArsenPidhoretskyi, line 31: For better readability, I suggest comparing as follows:
```
if not isinstance(animal, Herbivore):
    return

if animal.hidden is True:
    return

animal.health = animal.health - 50
if animal.health <= 0:
    Animal.alive.remove(animal)
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,5 +1,5 @@
 class Animal:
-    alive = []
+    alive: list = []
```
ArsenPidhoretskyi, line 2:
```suggestion
    alive: list[Animal] = []
```

ArsenPidhoretskyi (changes requested):

app/main.py
avkpol: this syntaxis results in "Unresolved reference 'Animal' "

ArsenPidhoretskyi: You can use:
`from __future__ import annotations`

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done

PR #304, opened 2023-01-18
solution

OleksandrGrk:


app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
```
ArsenPidhoretskyi, line 5: Provide typing

app/main.py
```diff
@@ -1 +1,39 @@
...
+            Animal.alive.remove(herbivore)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is True:
+            self.hidden = False
+        else:
+            self.hidden = True
```
ArsenPidhoretskyi, line 39: You can use:

ArsenPidhoretskyi (changes requested):

app/main.py
OleksandrGrk: I don`t see how to implement it? 
How self.hidden can be not self.hidden?

app/main.py
OleksandrGrk: Sorry, I can`t understand where do I need typing?) 
Do you mean from typing import Any or something like that?

ArsenPidhoretskyi: No just:
```suggestion
    alive: list[Animal] = []
```

app/main.py
OleksandrGrk: Done

app/main.py
OleksandrGrk: Done

app/main.py
```diff
@@ -1 +1,36 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Vasyl-Smutok, line 21: better avoid using `/`
You cat write like this
```
 return (
    f"{{Name: {self.name}, " 
    f"Health: {self.health}, " 
    f"Hidden: {self.hidden}}}"
)
```

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (approved):
Well done)

PR #308, opened 2023-01-22
'Solution'

andy77andy:
Hi, there) hope, everything is ok

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{Name: " + self.name + ", Health: " + str(self.health) + \
+               ", Hidden: " + str(self.hidden) + "}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbi: Herbivore) -> None:
+        if herbi in Animal.alive:
+            if herbi.hidden is False:
+                if isinstance(herbi, Herbivore):
+                    herbi.health -= 50
+                    if herbi.health <= 0:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,33 @@
...
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{Name: " + self.name + ", Health: " + str(self.health) + \
+               ", Hidden: " + str(self.hidden) + "}"
```
olena-hul, line 16: Use `()` to split lines, not `\`

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbi: Herbivore) -> None:
+        if herbi in Animal.alive:
+            if herbi.hidden is False:
+                if isinstance(herbi, Herbivore):
+                    herbi.health -= 50
+                    if herbi.health <= 0:
```
olena-hul, line 32: No need to do such a nesting. You can use 1 if to reduce animal's health, and after that, another separate if to remove from alive list

olena-hul (changes requested):

app/main.py
andy77andy: done

app/main.py
Vasyl-Smutok: please fix this like Olena says

Example:
```
        return ("Слава Україна!🇺🇦"
                "Героям Слава!🇺🇦")
```

Also use `f` string in this case

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {str(self.health)}," \
+               f" Hidden: {str(self.hidden)}}}"
```
olena-hul, line 17: Still `\` is used

olena-hul (changes requested):

vasylhnatiuk (approved):
Good job!

PR #309, opened 2023-01-23
Solution

xtina999:


app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+from typing import Any
+
+
+class Animal:
+    alive = []
```
ArsenPidhoretskyi, line 5: provide typing 
HINT: from __future__ import annotations
alive is List[Animal]

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> Any:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herb: Herbivore) -> Any:
```
ArsenPidhoretskyi, line 29:
```suggestion
    def bite(herbivore: Herbivore) -> Any:
```

ArsenPidhoretskyi (changes requested):

app/main.py
xtina999: why is it better to write this way? what is the usefulness of such writing?

app/main.py
```diff
@@ -1 +1,33 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
ArsenPidhoretskyi, line 18: Try to avoid '\' use brackets instead
```suggestion
        return (
            f"{{Name: {self.name},"
            f" Health: {self.health}," 
            f" Hidden: {self.hidden}}}"
        )
```

PR #312, opened 2023-01-25
Solution

Oleksy-Syrota:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+
+    alive = []
```
ArsenPidhoretskyi, line 3: provide type annotations here
HINT: use `from __future__ import annotations`

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, animal_herbivore: Herbivore) -> Herbivore:
+        if animal_herbivore.hidden is False\
+                and isinstance(animal_herbivore, Herbivore) is True:
+            animal_herbivore.health -= 50
+        for i in Animal.alive:
+            if i.health <= 0:
+                Animal.alive.remove(i)
```
ArsenPidhoretskyi, line 35: Why do you need a loop here? Due to this method only one animal can change hp

app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal_herbivore: Herbivore) -> Herbivore:
+        if animal_herbivore.hidden is False\
+                and isinstance(animal_herbivore, Herbivore) is True:
```
ArsenPidhoretskyi, line 31: rename `animal_herbivore` to `animal` or `herbivore`
also, `is False` and `is True` redundant in check

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,38 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> Herbivore:
+        if animal.hidden is False\
+                and isinstance(animal, Herbivore) is True:
```
andrewbelo, line 34:
```suggestion
        if not animal.hidden and isinstance(animal, Herbivore):
```

andrewbelo (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
```suggestion
    def __repr__(self) -> str:
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, "
            f"Hidden: {self.hidden}}}"
        )
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #317, opened 2023-01-30
Solution

MaxiTDI:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore):
+            if animal.hidden is False:
```
Grayder0152, line 36: `if not animal.hidden` - there is more correct solution, but this is not critical

Grayder0152 (approved):

PR #318, opened 2023-01-31
Solution

anastasiiashchoholieva:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if (isinstance(animal, Herbivore) is True
+           and animal.hidden is False):
```
andrewbelo, line 31:
```suggestion
        if isinstance(animal, Herbivore and not animal.hidden:
```

andrewbelo (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name},"\
+               f" Health: {self.health},"\
+               f" Hidden: {self.hidden}}}"
```
ArsenPidhoretskyi, line 18:
```suggestion
        return (
            f"{{Name: {self.name},"
            f" Health: {self.health},"
            f" Hidden: {self.hidden}}}"
        )
```

ArsenPidhoretskyi: Try to avoid '\' when it is possible

ArsenPidhoretskyi (changes requested):

stanhjr (approved):

PR #323, opened 2023-02-04
'Solution'

bohdan-yatsyna:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False
+                 ) -> None:
```
Vasyl-Smutok, line 9:
```suggestion
def __init__(
        self, 
        name: str, 
        health: int = 100, 
        hidden: bool = False
) -> None:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
```
Vasyl-Smutok, line 18:
```suggestion
return (
    f"{{Name: {self.name}, "
    f"Health: {self.health}, "
    f"Hidden: {self.hidden}}}"
)
```

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from __future__ import annotations
```
Vasyl-Smutok, line 1: Unnecessary import

Vasyl-Smutok (changes requested):

app/main.py
bohdan-yatsyna: Thank you for the remarks. Updated already.

ArsenPidhoretskyi (approved):
Well done!

PR #327, opened 2023-02-07
Developer

manjustice:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
+        if not isinstance(animal, Carnivore) and not animal.hidden:
+            animal.health -= 50
+        if animal.health <= 0:
+            Animal.alive.remove(animal)
+            del animal
```
MaybeTI, line 34: I'm sorry, I don't understand why you delete it again, maybe it ok, sorry

MaybeTI (approved):

app/main.py
manjustice: `Animal.alive.remove(animal)` here I delete it from list alive and `del animal` I delete object because it's dead

MaybeTI: Ok, good job!!!

Serhii-Leonenko, line 34: del is redundant because python garbage collector will do it for you)

app/main.py
```diff
@@ -1 +1,34 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
+        if not isinstance(animal, Carnivore) and not animal.hidden:
+            animal.health -= 50
+        if animal.health <= 0:
+            Animal.alive.remove(animal)
```
Serhii-Leonenko, line 33: The method should be static, self is redundant

Serhii-Leonenko (changes requested):

oigod (approved):

PR #330, opened 2023-02-07
1 commit

AndereLion:


app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
SafonovVladimir, line 6: Add indentation

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{" + f"Name: {self.name}, " \
+                     f"Health: {self.health}, " \
+                     f"Hidden: {self.hidden}" + "}"
```
SafonovVladimir, line 15: u can make f-string without concatenation

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Animal) -> None:
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health -= 50
+        if animal.health <= 0:
+            self.alive.remove(animal)
```
SafonovVladimir, line 24: I think, u use Class attribute "alive", so u should call it using Class.attribute. 
After that u don't use self in that method, so it will be staticmethod

SafonovVladimir (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+            Animal.alive.remove(animal)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
SafonovVladimir, line 36: Better  self.hidden = not self.hidden

SafonovVladimir (changes requested):

SafonovVladimir (approved):
GJ

app/main.py
Serhii-Leonenko: Move name to the newline to increase readability

app/main.py
```diff
@@ -1 +1,31 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Serhii-Leonenko, line 15: Do it without `\`

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+
```
Serhii-Leonenko, line 21: It's redundant an empty line here

Serhii-Leonenko (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

YuriiKindrat (approved):

PR #331, opened 2023-02-07
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,36 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return str(self.__dict__)
+
+    def __repr__(self) -> str:
+        return str(f"{{Name: {self.name}, "
+                   f"Health: {self.health}, "
+                   f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,35 @@
...
+        return str(self.__dict__).title().replace("'", "")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
manjustice, line 25: Better use here just `self.hidden = not self.hidden`

manjustice (changes requested):

app/main.py
SafonovVladimir: Cool! Thanks!
Done

manjustice (approved):

app/main.py
```diff
@@ -1 +1,32 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return str(self.__dict__).title().replace("'", "")
```
Serhii-Leonenko, line 16: interesting solution)

app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health.__le__(0):
```
Serhii-Leonenko, line 31: use operator `<=` instead of magic_method

Serhii-Leonenko (changes requested):

app/main.py
SafonovVladimir: Done

app/main.py
SafonovVladimir: more elegant than f-string

ramsty (approved):

YuriiKindrat (approved):

PR #332, opened 2023-02-08
Solution

Roman28101:


IvanRamyk (approved):

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive: list["Animal"] = []
```
IvanRamyk, line 2:
```suggestion
    alive: list[Animal] = []
```
type annotations should represented by classes, not strings

IvanRamyk: if you have problems with `type` is not defined, use
```
from __future__ imports annotations
```

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #334, opened 2023-02-08
Solution sry but flake8

NEkropo1:


DmytryTong (approved):
GJ. Code easy to read.

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = False
+        Animal.alive.append(self)
```
Serhii-Leonenko, line 8: Will it works if I want to create an instance with the `hidden = True` parameter?

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            "{"
+            + f"Name: {self.name}, "
+              f"Health: {self.health}, "
+              f"Hidden: {self.hidden}"
+            + "}"
+        )
```
Serhii-Leonenko, line 17: Use `{{}}` for that purpose instead of `+`

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herb: Herbivore) -> None:
+        if type(herb) is Herbivore and not herb.hidden:
```
Serhii-Leonenko, line 28: I think it's better to use isinstance() method here for this purpose

Serhii-Leonenko (changes requested):

app/main.py
NEkropo1: flak8 (

app/main.py
NEkropo1: > I think it's better to use isinstance() method here for this purpose

what if we inherit class from carnivore?

app/main.py
NEkropo1: > Use `{{}}` for that purpose instead of `+`

When I made that and checked with black it made one string which was > 79, so I needed to concatenate strings, that's why it's easier to see logics of concatenation...

app/main.py
NEkropo1: > Will it works if I want to create an instance with the `hidden = True` parameter?

Sure not, but I wasn't sure either we going to inherit it and change only for rare classes, how it should be in ideal case(Which animal takes hidden by default?) or it should be hidden, and since it passed tests, I was sure I made it right way, as said in README.MD -> according to nature's logic. 
sry )

app/main.py
NEkropo1: and as for me, explicit is better than implicit, so when we explicitly add concatenation, it's more obvious to other programs

NickSavchyn (approved):

PR #337, opened 2023-02-09
Solution

WDemidenko:


app/main.py
```diff
@@ -1 +1,33 @@
...
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Herbivore) -> None:
```
oksanaaam, line 25: you can make this function statical, but it's not necessary

app/main.py
```diff
@@ -1 +1,33 @@
...
+        if isinstance(other, Herbivore) and other.hidden is False:
+            other.health -= 50
+            if other.health <= 0:
+                Animal.alive.remove(other)
+
+# lion = Carnivore("Simba")
+# len(Animal.alive) == 1
+# isinstance(Animal.alive[0], Carnivore) is True
```
oksanaaam, line 33: also you better remove these comments

app/main.py
WDemidenko: In the task there was an example of using this function "lion.bite(rabbit)", so I didn't make it static

app/main.py
WDemidenko: thanks

oksanaaam (approved):

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
```
Serhii-Leonenko, line 8: Fix indentation here to increase readability, move name to new line

app/main.py
```diff
@@ -1 +1,30 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Serhii-Leonenko, line 16: Do it without `\`

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
```
Serhii-Leonenko, line 26: I think other is not the best name for the parameter in this case

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
```
wQuelS, line 28: animal is Animal here. It may be also a `Carnivore`

wQuelS (changes requested):

wQuelS (approved):
Great

PR #340, opened 2023-02-09
Solution

yura-hudzovskyi:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def check_alive(animal: Animal) -> None:
+        if animal.health <= 0:
+            Animal.alive.remove(animal)
+
+
+class Herbivore(Animal):
+    @staticmethod
+    def hide() -> None:
+        for animal in Animal.alive:
+            animal.hidden = not animal.hidden
```
Zenionoi, line 33: `Changes the hidden property of the beast`, not for all animals

Zenionoi (changes requested):

viktoria-rybenchuk (approved):

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    @staticmethod
+    def check_alive(animal: Animal) -> None:
+        if animal.health <= 0:
+            Animal.alive.remove(animal)
```
Serhii-Leonenko, line 26: It's better to add self parameter to the method, static is not a good idea in this case

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if not animal.hidden and not isinstance(animal, Carnivore):
+            animal.health -= 50
+        Animal.check_alive(animal)
```
YuriiKindrat, line 36: no need to use this method as class method, u can use this method for bitten animal

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #344, opened 2023-02-09
Solution

Zenionoi:


app/main.py
```diff
@@ -1 +1,35 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        tp = self
+        return f"{{Name: {tp.name}, Health: {tp.health}, Hidden: {tp.hidden}}}"
```
oksana0410, line 18: not sure if tp is useful here. maybe to can skip this step and use self.name

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(victim: Herbivore | Carnivore) -> None:
+        if isinstance(victim, Herbivore):
+            if not victim.hidden:
```
oksana0410, line 32: not sure if it is better, but maybe you can make one statement with isinstance and not victim in 1 line

oksana0410 (approved):

app/main.py
Zenionoi: I change self.name to t(em)p.name for no tearing of the tape and have length less than the maximum of flake8

app/main.py
Zenionoi: Thanks for the comments, it's really better to use that, but i don't want to combine the two `not statement`

Serhii-Leonenko: Yes, it's better to make the condition in one line with operator `and` instead of nested `if`

app/main.py
Serhii-Leonenko: Yes, it's a redundant variable. Just use brackets to wrap f-string according to flake8

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(victim: Herbivore | Carnivore) -> None:
```
YuriiKindrat, line 30: fix type annotation here, Carnivore can't be bitten 
image

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #347, opened 2023-02-09
solution

amber-marichi:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        health: int = 100,
+        hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        me = f"Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}"
```
Morty67, line 19: why not used return f"Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}" ??

app/main.py
```diff
@@ -1 +1,35 @@
...
+        return "{" + me + "}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = bool(int(self.hidden) - 1)
```
Morty67, line 26: you can do it easier

Morty67 (changes requested):

Morty67 (approved):

app/main.py
```diff
@@ -1 +1,35 @@
...
+        hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
YuriiKindrat, line 20: try to avoid using " \ " u can use brackets here, to break string

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #348, opened 2023-02-09
Solution

ghost:


amber-marichi (approved):
looks cool. approved

app/main.py
```diff
@@ -1 +1,31 @@
...
+        hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return ("{Name: %s, Health: %s, Hidden: %s}"
+                % (self.name, self.health, self.hidden))
```
Serhii-Leonenko, line 17: Make it with f-string, it has better readability

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
```
Serhii-Leonenko, line 27: I think other is not the best name in this case

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
wQuelS, line 28: I guess `Animal` instance is passing in this method, not a specific kind of it like `Herbivore`. That's why you're doing a type check one line below

wQuelS (changes requested):

YuriiKindrat (approved):

PR #350, opened 2023-02-10
'Solution'

Morty67:


app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore):
```
AlessaQ, line 29: For my opinion:
Here if you create a new class Fish, carnivore cannot demage fish, cause she is not Herbivore. So better to use `if not` same class.

AlessaQ (changes requested):

wQuelS (approved):
Great

PR #352, opened 2023-02-10
Develop

MikhailLyvak:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, \
+            Health: {self.health}, \
+            Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
```
Yulia-Mala, line 26: pls check in the check list (under the readme text) how to write it in one line using not keyword

app/main.py
```diff
@@ -1 +1,39 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, \
```
Yulia-Mala, line 20: one test fails, I guess because on this method. I am not sure about the reason but in my case pycharm has added additional " symbols on the end of the line when put text on the new line
        return f"{{Name: {self.name}, " \
                f"Health: {self.health}, " \
                f"Hidden: {self.hidden}}}"

Yulia-Mala (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
Serhii-Leonenko, line 22: Do it without `\`

app/main.py
Serhii-Leonenko: Yes, you should fix it, read the checklist

Serhii-Leonenko (changes requested):

ada-krav (approved):
GJ

YuriiKindrat (approved):

PR #354, opened 2023-02-10
Solution

ZhAlexR:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> None:
+        if not other.hidden and isinstance(other, Herbivore):
+            other.health -= 50
+            other.is_alive()
```
Anatolii-Poznyak, line 34: > `other.is_alive()`

Nice ! I wish I had thought of that.

app/main.py
```diff
@@ -1 +1,34 @@
...
+        Animal.alive.append(self)
+
+    def is_alive(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Anatolii-Poznyak, line 20: I also searched for "how to split up a long f-string". Actually you can do it w/o that "\" symbol. Just use () after return. Looks much cleaner 
Screenshot 2023-02-10 at 16 06 44

Anatolii-Poznyak (approved):

app/main.py
ZhAlexR: @Anatolii-Poznyak, thanks! Looks really better! 
Lets wait for mentor comment about it.

YuriiKindrat: yeah, it is better to avoid " \ " symbol

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #357, opened 2023-02-10
'Solution'

NazarHladaniuk:


inrock1 (approved):
All right

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(unlucky: Animal) -> None:
```
wQuelS, line 31: `animal` would be better name, I guess

wQuelS (changes requested):
Really nice work, just one small comment

PR #358, opened 2023-02-10
Solution

TeliksPG:


app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
skmiwa1, line 9: indentation wrong
")" on another line

skmiwa1 (changes requested):
line 9

app/main.py
ProdyRodion: @skmiwa1  if self on next line - yes, but not in this situation

skmiwa1: name on the next line! line 7

skmiwa1 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 19: try not to use " \ " here, u can use brackets for breaking line

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Carnivore) -> None:
```
YuriiKindrat, line 29: are u sure about Carnivore in type annotation ?

app/main.py
```diff
@@ -1 +1,33 @@
...
+        if not animal.hidden and isinstance(animal, Herbivore):
+            animal.health -= 50
+        if animal.health < 1:
+            del Animal.alive[Animal.alive.index(animal)]
```
YuriiKindrat, line 33: simplify this line. We access class attributes through class name and it's better to use remove when value is given and del when an index is given. You do unnecessary work here

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #365, opened 2023-02-10
Solution

ostapT:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{" \
+               f"Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}" \
+               f"}}"
+
```
wQuelS, line 18: try to avoid line breaks

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
wQuelS, line 25: consider using `self.hidden = not self.hidden`

wQuelS (changes requested):

wQuelS (approved):
Really nice

PR #367, opened 2023-02-10
'Solution'

alexlega:


ostapT (approved):
GJ!

app/main.py
```diff
@@ -1 +1,32 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}, Hidden: {self.hidden}}}"
+
```
YuriiKindrat, line 18: try not to use " \ " here, u can break the string using brackets

app/main.py
```diff
@@ -1 +1,32 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbiv_obj: Herbivore) -> None:
```
YuriiKindrat, line 28: herbiv_obj is a bad name for variable, use full names for naming

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #369, opened 2023-02-10
Solution

tytomyr:


app/main.py
```diff
@@ -1 +1,40 @@
...
+                 health: int = 100,
+                 hidden: bool = hidden) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}, Hidden: {self.hidden}}}"
```
wQuelS, line 15: try to avoid those line breaks

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Carnivore(Animal):
+
+    def carnivore(self, name: str) -> None:
+        super().__init__(name)
```
wQuelS, line 21: huh? do you really need it ?

app/main.py
```diff
@@ -1 +1,40 @@
...
+            beast.health -= 50
+        if beast.health <= 0:
+            Animal.alive.remove(beast)
+
+
+class Herbivore(Animal):
+
+    def herbivore(self, name: str) -> None:
+        super().__init__(name)
+
```
wQuelS, line 35: same

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
wQuelS, line 40: I guess `self.hidden = not self.hidden` is more reliable

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+            beast.health -= 50
+        if beast.health <= 0:
+            Animal.alive.remove(beast)
+
+
+class Herbivore(Animal):
+
+    def herbivore(self, name: str) -> None:
+        pass
+
```
Nattalli, line 38: I believe you can remove this method at all

app/main.py
```diff
@@ -1 +1,43 @@
...
+        name = self.name
+        health = self.health
+        hide = self.hidden
+        return f"{{Name: {name}, Health: {health}, Hidden: {hide}}}"
+
+
+class Carnivore(Animal):
+
+    def carnivore(self, name: str) -> None:
+        pass
```
Nattalli, line 24: and this also you can remove

app/main.py
Nattalli: I will explain:
```
self.hidden = not self.hidden
```
if self.hidden = True then we will make it False,
if self.hidden = False then we will make it True

You can replace 4 lines in 1 that @wQuelS recommended

Nattalli (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #374, opened 2023-02-11
Solution

me1nyk:


app/main.py
```diff
@@ -1 +1,33 @@
...
+                 ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
oksanaaam, line 19: you need write without brackets, but its not big issue

oksanaaam (approved):

app/main.py
wQuelS: yeah, but the big issue are line breaks:)

wQuelS (changes requested):

app/main.py
me1nyk: I have some issue with PyCharm, spent a lot of time with is, will ask someone tomorrow what i need to do with is, probably during QnA.
Please check second commit (posted after first one).

me1nyk: issue resolved, i had an issue with synchronization of local and remote branches
please check last commit, it is final solution

app/main.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
```
Nattalli, line 10: btw, it's a bad code style.
Good:
```python
("a", "b", "c", "d")
```
Also good:
```python
(
"a",
"b",
"c",
"d"
)
```

Bad:
```python
(
"a", "b"
"c", "d"
)
```
Also bad:
```python
("a",
"b",
"c",
"d"
)
```

Nattalli (changes requested):

app/main.py
me1nyk: thanks, thats look batter)

Nattalli (approved):

PR #375, opened 2023-02-11
Solution

myroslavatrschk:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
```
manjustice, line 12: Better way use attribute alive as a Animal's attribute

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if not self.hidden else False
```
manjustice, line 22: here u can use just `not self.hidden`

manjustice (changes requested):

manjustice (approved):

YuriiKindrat (approved):

PR #379, opened 2023-02-12
Solution

destr34:


oksanaaam (approved):

app/main.py
```diff
@@ -1 +1,37 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".\
+            format(self.name, self.health, self.hidden)
```
wQuelS, line 15: avoid line breaks.
use `f-strings`

app/main.py
```diff
@@ -1 +1,37 @@
...
+            Animal.alive.pop()
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
wQuelS, line 29: consider using `self.hidden = not self.hidden`

app/main.py
```diff
@@ -1 +1,37 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".\
+            format(self.name, self.health, self.hidden)
+
+    @staticmethod
+    def is_dead(lunch_name: str) -> None:
+        if lunch_name.health <= 0:
+            Animal.alive.pop()
```
wQuelS, line 20: no need to make it `static`. 
use `remove` instead of `pop`

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
YuriiKindrat, line 7: fix indentation here, one arg per line, for better code style

app/main.py
```diff
@@ -1 +1,35 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 16: try not to use " \ " use brackets instead, for better code style

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, lunch_name: str) -> None:
```
wQuelS, line 33: first: do u even use this `self` ? consider making this method as static.
second: lunch_name is a bad variable name, and it is not a `str`

app/main.py
```diff
@@ -1 +1,36 @@
...
+        if isinstance(lunch_name, Herbivore) and lunch_name.hidden is False:
```
wQuelS, line 34: `is False` is not actually needed, since `if something` statement returns boolean value, you can mix it with `not` and that's it

wQuelS (changes requested):

wQuelS (approved):
Good job!

PR #380, opened 2023-02-12
'Solution'

AnnaBoichenko:


yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name_str = f"Name: {self.name}, "
+        health_str = f"Health: {self.health}, "
+        hidden_str = f"Hidden: {self.hidden}"
+        return "{" + name_str + health_str + hidden_str + "}"
```
YuriiKindrat, line 18: no need to create this variables here, use brackets to break the line

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(obj: Herbivore) -> None:
+        if isinstance(obj, Herbivore) and not obj.hidden:
```
YuriiKindrat, line 31: obj is bad name for variable, use f.e. use animal or smth like that

YuriiKindrat (changes requested):

app/main.py
AnnaBoichenko: but otherwise, it will come this way, which was in bad practice examples
image

PR #381, opened 2023-02-13
'Solution'

RuslanLopatin:


MikhailLyvak (changes requested):
image
в тебе якісь технічні чьоколадкі тут. Думаю треба пофіксити

app/main.py
```diff
@@ -1 +1,34 @@
...
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.health = 100
+        self.hidden = False
+        self.__class__.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 13: try not to use " \ " use brackets instead for better code style

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def die(self) -> None:
+        self.__class__.alive.remove(self)
```
YuriiKindrat, line 16: u can just use class name here, to write less code
also u can make this method as classmethod

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: bool) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
```
YuriiKindrat, line 22: no need to use blank return here, u can just make a condition to use the code below

app/main.py
```diff
@@ -1 +1,34 @@
...
+        herbivore.health -= 50
+        if herbivore.health <= 0:
+            herbivore.health = 0
+            herbivore.die()
+
+
+class Herbivore(Animal):
+    def herb(self, name: str) -> None:
+        super().__init__(name)
```
YuriiKindrat, line 31: why do u need this method here ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+                herbivore.die()
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+    def die(self) -> None:
+        self.__class__.alive.remove(self)
```
YuriiKindrat, line 40: no need to writ self.class.alive.. u can just call it like this Animal.alive.re... for better code style

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def die(cls, animal: str) -> None:
+        cls.alive.remove(animal)
```
YuriiKindrat, line 22: why do u need this method, if u have such method in Herbivore class ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #383, opened 2023-02-13
'Solution'

iishchenko:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        self.hidden = hidden
+        if health > 0:
+            self.alive.append(self)
+
+    def __repr__(self) -> None:
+        return "{"\
+            + f"Name: {self.name}," \
+              f" Health: {self.health}," \
+              f" Hidden: {self.hidden}"\
+            + "}"
```
Vasyl-Smutok, line 25: better to use `()` then `\`
```suggestion
        return (
             f"{{Name: {self.name}, "
             f"Health: {self.health}, "
             f"Hidden: {self.hidden}}}"
         )
```

Vasyl-Smutok (changes requested):

iishchenko:
Changes:
1)better to use () than \
2)if redundant here

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore) and not isinstance(animal, Carnivore):
+            if not animal.hidden:
+                animal.health -= 50
+                if animal.health < 0:
```
ArsenPidhoretskyi, line 40: Maybe write here <=
and the move animal.alive.remove(animal) to this part?

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore) and not isinstance(animal, Carnivore):
```
ArsenPidhoretskyi, line 37: and not isinstance(animal, Carnivore) is not needed here it is always false

app/main.py
```diff
@@ -1 +1,43 @@
...
+            if not animal.hidden:
```
ArsenPidhoretskyi, line 38: Write with and this statement

ArsenPidhoretskyi (changes requested):

iishchenko:
Changes:
1)Maybe write here <=
and the move animal.alive.remove(animal) to this part?
2)and not isinstance(animal, Carnivore) is not needed here it is always false
3)Write with and this statement

katevasilenko (approved):
Looks good to me

PR #384, opened 2023-02-14
Solution

dogmoony:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
oksanaaam, line 18: these brackets are issue, you need delete it

oksanaaam (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        result = f"Name: {self.name}, \
+Health: {self.health}, \
+Hidden: {self.hidden}"
+        return "{" + result + "}"
```
Serhii-Leonenko, line 19: Remake this, you can use brackets and remove `\`. And you can use `{{}}` in your f-string to display it like { }

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, obj: Animal) -> None:
+        if obj.hidden is False and isinstance(obj, Herbivore):
+            obj.health -= 50
+            if obj.health <= 0:
+                Animal.alive.remove(obj)
```
Serhii-Leonenko, line 34: Use more descriptive name instead of `obj`

Serhii-Leonenko (changes requested):
Fix your code according to the comments

app/main.py
dogmoony: what brackets???

app/main.py
wQuelS: still not fixed

wQuelS (changes requested):

app/main.py
dogmoony: I don't understand how to do that

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Animal) -> None:
```
wQuelS, line 28: do u use this `self` somehow? maybe consider using staticmethod here ?

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if animal.hidden is False and isinstance(animal, Herbivore):
```
wQuelS, line 29: `if animal` or `if not animal` -> both statements return a boolean value, so no need to write `is False`

wQuelS (changes requested):

PR #385, opened 2023-02-14
Solution

lotrik84:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
```
ArsenPidhoretskyi, line 17: Try to avoid usage of `'\'`  Using brackets is better)

ArsenPidhoretskyi (changes requested):

app/main.py
lotrik84: This is not me) PyCharm did it)

ArsenPidhoretskyi (approved):
Well done!

PR #390, opened 2023-02-20
Solo

IgoraStar:


app/main.py
```diff
@@ -1 +1,33 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.__class__.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
Grayder0152, line 16: Pls use `()` instead `\`

app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.__class__.alive.append(self)
```
Grayder0152, line 13: Use `self.alive.append(self)`

Grayder0152 (changes requested):

app/main.py
Paul-Starodub: Why? Alive is attribute of class

Vasyl-Smutok: better like this 
```
Animal.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
```
Vasyl-Smutok, line 18:
```suggestion
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, "
            f"Hidden: {self.hidden}}}"
        )
```

app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
```
Vasyl-Smutok:
```suggestion
        Animal.alive.append(self)
```

Vasyl-Smutok (changes requested):

app/main.py
Paul-Starodub: Hi) it is hard coding((( If we change name of class we will have an error Better is self.__class__.alive.append(self) .IMHo

ArsenPidhoretskyi (approved):
Well done!

PR #398, opened 2023-02-27
Solution

kostya-kononenko:


app/main.py
```diff
@@ -1 +1,30 @@
...
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
vasylhnatiuk, line 13: use `() ` for long `str` instead of `\`

vasylhnatiuk (changes requested):

app/main.py
kostya-kononenko: I do not understand how? Can you post an example?

Grayder0152 (approved):

Abnormaltype (approved):

PR #399, opened 2023-02-28
Solution_for_herbivores-and-carnivores

AndriiFedoryshyn:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
+    def bite(self, hernivore: Herbivore) -> None:
+        if isinstance(hernivore, Herbivore) and hernivore.hidden is False:
+            hernivore.health -= 50
+        for animal in Animal.alive:
+            if animal.health <= 0:
```
ArsenPidhoretskyi, line 34: Not fully logical. Due to this method, only one animal can have an hp less than 0. Other is not possible

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
```
olena-hul, line 35: No need to check every animal. Check only hernivore.health

olena-hul (changes requested):

PR #403, opened 2023-03-08
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False)\
```
Grayder0152, line 7: do not use `\`

app/main.py
```diff
@@ -1 +1,32 @@
...
+            -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{" + "Name: " + self.name + ", Health: " + \
```
Grayder0152, line 15: use `()` instead of `()`

Grayder0152 (approved):

app/main.py
V-Shkrobatskyi: same brackets?

V-Shkrobatskyi: what do yo mean?

PR #407, opened 2023-03-16
Solution

MaksymUK:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(creature: Animal) -> None:
+        if isinstance(creature, Carnivore) or creature.hidden is True:
+            pass
+        else:
+            creature.health -= 50
+            creature.check_health()
```
Bohdan-Salamakha, line 47: u don't need to do comparison `creature.hidden is True` just `creature.hidden`
use "not" keyword to simplify the condition (remove "pass" and "else")

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #411, opened 2023-03-20
'Solution#GeeGee'

Jhaivir:


app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if isinstance(victim, Herbivore) and not victim.hidden:
+            victim.health -= 50
+        for animal in range(len(Animal.alive)):
+            if Animal.alive[animal].health <= 0:
+                del Animal.alive[animal]
```
Grayder0152, line 40: it is not necessary, check this only for `victim`

app/main.py
```diff
@@ -1 +1,40 @@
...
+            + "}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
Grayder0152, line 30: `self.hidden = not self.hidden`

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
Grayder0152, line 13:
```suggestion
        self.alive.append(self)
```

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,38 @@
...
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            "{"
+            + f"Name: {self.name}, "
+            + f"Health: {self.health}, "
+            + f"Hidden: {self.hidden}"
+            + "}"
```
ArsenPidhoretskyi, line 21: Is really `+` needed here?

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if isinstance(victim, Herbivore) and not victim.hidden:
+            victim.health -= 50
+        if victim.health <= 0:
+            for animal in range(len(Animal.alive)):
+                if Animal.alive[animal] is victim:
+                    del Animal.alive[animal]
```
ArsenPidhoretskyi, line 38: Seriously, it is possible that another animal than victim can have hp <= 0 due to this method?

ArsenPidhoretskyi (changes requested):

app/main.py
Jhaivir: It's not possible, but I need to remove the dead victim from the list, and for this I need to find it, so I'm looking for it by the index

app/main.py
Jhaivir: Otherwise, flake8 complains about the long string.

app/main.py
katevasilenko: Just use `.remove()` method and victim as an argument

katevasilenko (changes requested):
Also you’ve shouldn’t push your PR to master branch. Please fix this

app/main.py
Jhaivir: Wow, that's the method I missed, thanks

Vasyl-Smutok (approved):
Good Job 👍

PR #414, opened 2023-03-25
class Animal done

dirolius:


app/main.py
```diff
@@ -1 +1,31 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and herbivore.hidden is False:
```
Bohdan-Salamakha, line 28: use "is" only to compare value with None
use `and not herbivore.hidden`

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #415, opened 2023-03-25
Solution

DmytroBondariev:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{"\
+               f"Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}"\
+               "}"
```
Vasyl-Smutok, line 23: better to use `()` then `\`

app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from __future__ import annotations
```
Vasyl-Smutok, line 1: Unnecessary import

app/main.py
```diff
@@ -1 +1,39 @@
...
+               f"Hidden: {self.hidden}"\
+               "}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
Vasyl-Smutok, line 31: try to write it in one line 

HINT:
use `not`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+            f"Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}"
+            "}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is not True else False
```
Vasyl-Smutok, line 27:
```suggestion
        self.hidden = not self.hidden
```

Vasyl-Smutok (changes requested):

app/main.py
Dmytry95: а это интересно)

katevasilenko (approved):
Looks good to me

PR #418, opened 2023-03-28
Solution

Maksym-Filipchuk:
How can I make last part of the test 
a have already tried     def __str__(self) -> str:
        return str([
            {"Name": animal.name, "Health": animal.health,
             "Hidden": animal.hidden}
            for animal in Animal.alive
        ])
and 
def __str__(self):
but still can`t solve it

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> any:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return str([
+            {"Name": animal.name, "Health": animal.health,
+             "Hidden": animal.hidden}
+            for animal in Animal.alive
+        ])
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> any:
+        if not other.hidden and isinstance(other, Herbivore):
+            other.health -= 50
+        if other.health <= 0:
+            Carnivore.alive.remove(other)


app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return str([
+            {"Name": animal.name, "Health": animal.health,
+             "Hidden": animal.hidden}
+            for animal in Animal.alive
+        ])
```
ArsenPidhoretskyi, line 17: Maybe need to return string representation for one animal, not for all alive?

ArsenPidhoretskyi: And also not str, you need use repr

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
```
ArsenPidhoretskyi, line 23: Why need to return self.hidden??

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> any:
+        if not other.hidden and isinstance(other, Herbivore):
+            other.health -= 50
+        if other.health <= 0:
+            Carnivore.alive.remove(other)
```
ArsenPidhoretskyi, line 32: Better to use Animal.alive

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
------

app/main.py
```diff
@@ -1 +1,28 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
```
ArsenPidhoretskyi, line 18: return type annotations is wrong

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> any:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
katevasilenko, line 14: It’s better to use brackets to avoid \ symbol

katevasilenko (approved):
Looks good to me. It would great if you provide small code style change

PR #419, opened 2023-03-28
task done

chipapapo:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive += [self]
```
ArsenPidhoretskyi, line 12: Animal.alive.append(self)

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, prey: Herbivore) -> None:
+        if isinstance(prey, Herbivore):
+            if not prey.hidden:
```
ArsenPidhoretskyi, line 31: Combine into one condition this, it unlogical to do in two

PR #426, opened 2023-03-30
HERBIVORES AND CARNIVORES

volodinsy:


vasylhnatiuk (changes requested):
image
fix flake8 tests

PR #436, opened 2023-04-08
Solution

Munkoor:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is True:
+            self.hidden = False
+        else:
+            self.hidden = True
```
ArsenPidhoretskyi, line 25: You can simplify this by using not:
self.hidden = not self.hidden

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def check_is_alive(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
Vasyl-Smutok, line 17: better to use `()` then `\`
```suggestion
        return (
            f"{{Name: {self.name}, Health: {self.health},"
            f" Hidden: {self.hidden}}}"
        )
```

Vasyl-Smutok (changes requested):

vasylhnatiuk (approved):
Good job!

PR #437, opened 2023-04-08
Solution

flatchers:
I don't understand, what must be in annotation in magic method of constructor __init__ and in last fuction annotation ofvariable?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,34 @@
-# write your code here
+class Animal:
+    alive: list = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: classmethod) -> str:
```
ArsenPidhoretskyi, line 30:
```suggestion
    def bite(self, herbivore: Herbivore) -> str:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
```
ArsenPidhoretskyi, line 26: No need to return this

ArsenPidhoretskyi (changes requested):

flatchers (commented):
deletes and changes

PR #451, opened 2023-04-17
Solution_develop

sbilikepy:


app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+class Animal(object):
+    alive = []
+
+    def __init__(self, name: str, *args: int) -> None:
```
ArsenPidhoretskyi, line 4: it is wrong solution. It must accept 3 arguments:
name: str, hidden: bool, heath: int
But hidden and heath must have default argument

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if len(args) != 0:
+            self.health = args[0]
+        else:
+            self.health = 100
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return f"\nName = {self.name},\n" \
+               f"Current hp = {self.health},\n" \
+               f"Hiding? = {'Yes' if self.hidden == True else 'No'}\n"
```
ArsenPidhoretskyi, line 16: Use brackets instead of \
```suggestion
        return ("\nName = {self.name},\n" 
               f"Current hp = {self.health},\n" 
               f"Hiding? = {'Yes' if self.hidden == True else 'No'}\n"
           )
```

ArsenPidhoretskyi: Fix in repr also

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self.health <= 0:
+            Animal.alive.remove(self)
+        return self.health
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore) -> None:
+        if isinstance(target, Herbivore) and target.hidden is False:
+            target.__isub__(50)
```
ArsenPidhoretskyi, line 39:
```suggestion
            target -= 50
```
must be. If you use magic methods use them properly

app/main.py
```diff
@@ -1 +1,39 @@
...
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+    def __isub__(self, damage: int) -> int:
```
ArsenPidhoretskyi, line 28:
```suggestion
    def __isub__(self, damage: int) -> Animal:
```

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.health -= damage
+        if self.health <= 0:
+            Animal.alive.remove(self)
+        return self.health
```
ArsenPidhoretskyi, line 32:
```suggestion
        return self
```

ArsenPidhoretskyi (changes requested):

app/main.py
sbilikepy: Сould i use dunder method in this way:
```
def __isub__(self, damage: int = 50) -> int:
    self.health -= damage
    if self.health <= 0:
        Animal.alive.remove(self)
    return self.health
```
+
`target.__isub__()
`
or the functionality of the method should be **limited to the purpose** of the method (subtraction in this case)?

sbilikepy: or even:

```
def __isub__(self) -> Animal:
    self.health -= 50
    if self.health <= 0:
        Animal.alive.remove(self)
    return self
```

vasylhnatiuk (approved):

PR #452, opened 2023-04-18
Solution of animal

svyshcho:


app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
+        else:
```
ArsenPidhoretskyi, line 33: No need to use else: because we already return in other case

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def dead_animal(self) -> None:
```
ArsenPidhoretskyi, line 13: dead_animal rename to dead. Because we already know that it related to animal

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
```
ArsenPidhoretskyi, line 31:
```suggestion
        if not isinstance(herbivore, Herbivore) or herbivore.hidden:
```
Because if someone pass an Animal?

ArsenPidhoretskyi (changes requested):

app/main.py
svyshcho: Done

app/main.py
svyshcho: Done

app/main.py
svyshcho: Done

olena-hul:
Good job!

PR #456, opened 2023-04-21
Solution

KittNeKit:


app/main.py
```diff
@@ -1 +1,27 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore) and animal.hidden is False :
```
Bohdan-Salamakha, line 24:
```suggestion
        if isinstance(animal, Herbivore) and not animal.hidden:
```
u don't need to use is comparison here

app/main.py
```diff
@@ -1 +1,27 @@
...
+    def __init__(self, name: str, health: int = 100,) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Bohdan-Salamakha, line 13:
```suggestion
        return (
        f"{{Name: {self.name}, "
               f"Health: {self.health}, "
               f"Hidden: {self.hidden}}}"
               )
```

Bohdan-Salamakha (changes requested):
v1

PR #466, opened 2023-04-28
Solution

neostyle88:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def __repr__(self) -> str:
+        return (
+            "{Name: "
+            + self.name
+            + ", Health: "
+            + str(self.health)
+            + ", Hidden: "
+            + str(self.hidden)
+            + "}"
+        )
```
Bohdan-Salamakha, line 21: use f-strings and triple-quotes

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        elif self.hidden is True:
+            self.hidden = False
```
Bohdan-Salamakha, line 29:
```suggestion
        self.hidden = not self.hidden
```

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, victim: Animal) -> None:
+        if victim.hidden is False and isinstance(victim, Herbivore):
```
Bohdan-Salamakha, line 34: u don't need to use id comparison in this case
```suggestion
        if not victim.hidden and isinstance(victim, Herbivore):
```

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,30 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, Health: "
+            f"{str(self.health)}, "
+            f"Hidden: {str(self.hidden)}}}"
```
Bohdan-Salamakha, line 16: u don't need to use str()
every value will be converted into str automatically in the f-string

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #470, opened 2023-04-30
Develop2

melnikalex1977:


melnikalex1977:
1

melnikalex1977 (commented):
1

app/main.py
```diff
@@ -1 +1,29 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: type[Herbivore]) -> None:
```
ArsenPidhoretskyi, line 24: It does not accept type
It accept instance

ArsenPidhoretskyi (changes requested):

PR #475, opened 2023-05-04
Py Herbivores and carnivores

alexgubarev88:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def is_alive(self) -> None:
+        if self.health <= 0:
+            self.health = 0
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
```
vasylhnatiuk, line 19: use () for long str instead of \

PR #484, opened 2023-05-15
solution

IvanKorshunovE:


app/main.py
```diff
@@ -1 +1,40 @@
...
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, Hidden: {self.hidden}}}"
```
ArsenPidhoretskyi, line 20: Use brackets instead of \
```suggestion
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, Hidden: {self.hidden}}}"
        )
```

app/main.py
```diff
@@ -1 +1,40 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(prey: Type[Herbivore]) -> None:
```
ArsenPidhoretskyi, line 35: it accepts instance Herbivore

ArsenPidhoretskyi (approved):
Well done!

PR #489, opened 2023-05-18
'Solution'

ilizh:


app/main.py
```diff
@@ -1 +1,27 @@
...
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
EdAlekseiev, line 13: Use `()` to split lines

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
Bohdan-Salamakha, line 5:
```suggestion
    def __init__(
                 self,
                 name: str,
                 health: int = 100,
                 hidden: bool = False
                 ) -> None:
```

app/main.py
```diff
@@ -1 +1,29 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: bool) -> list:
```
Bohdan-Salamakha, line 24: invalid typehint for herbivore

add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #492, opened 2023-05-22
Herbivores and carnivores task complete

thorium-zero:


app/main.py
```diff
@@ -1 +1,50 @@
...
+            return
+        self._health -= value
+        if self._health < 1:
+            Animal.alive.remove(self)
+
+
+class Carnivore(Animal):
+    def bite(self, prey: Herbivore) -> None:
+        if not isinstance(prey, Herbivore) or prey.hidden:
+            return
```
YuriiKindrat, line 49: no need to use blank return here, u can use backward condition for this

YuriiKindrat (changes requested):

app/main.py
thorium-zero: what's wrong with empty returns? oO

LiudmylaKulinchenko (changes requested):
Remove the empty line and rewrite without protected arguments, please 🌺 You can read an article about it: [The Meaning of Underscores in Python](The Meaning of Underscores in Python) 🐥

app/main.py
```diff
@@ -1 +1,53 @@
...
+from __future__ import annotations
+
+
+class Animal:
+    alive: list[Animal] = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 is_hidden: bool = False) -> None:
+        self._name: str = name
```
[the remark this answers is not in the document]
thorium-zero: I can't do that. Not in this lifetime at least :D

LiudmylaKulinchenko (approved):
Good job 🍬🔥🚀

PR #501, opened 2023-05-26
finish lab

XOctOpus1:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
ArsenPidhoretskyi, line 17: Better to use brackets instead of \
```suggestion
        return (f"{{Name: {self.name}, "
               f"Health: {self.health}, "
               f"Hidden: {self.hidden}}}")
```

ArsenPidhoretskyi (changes requested):

app/main.py
XOctOpus1: Flake8 have an error 
E128 continuation line under-indented for visual indent

ArsenPidhoretskyi (approved):
Well done!

PR #505, opened 2023-06-01
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        self.alive.append(self)
```
YuriiKindrat, line 8: u should use class name for getting class attribute

app/main.py
```diff
@@ -1 +1,29 @@
...
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
YuriiKindrat, line 21: try to simplify this, u can make it in one line 
hint: use `not`

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            if other.health <= 0:
+                other.alive.remove(other)
```
YuriiKindrat, line 29: u should use class name here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #508, opened 2023-06-01
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1 +1,38 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
YuriiKindrat, line 16: it is better to use class name for getting class attribute

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore | Carnivore) -> None:
```
YuriiKindrat, line 33: could u use parent class for type annotation here?

PR #509, opened 2023-06-01
Solution

imelnyk007:


app/main.py
```diff
@@ -1 +1,29 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 13: try to avoid using linebreakers, u can use brackets instead

YuriiKindrat (changes requested):

PR #512, opened 2023-06-01
solved

oleksiikolii:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbi: Herbivore) -> None:
+        if isinstance(herbi, Herbivore) and not herbi.hidden:
```
YuriiKindrat, line 29: herbi is quite cute name for an instance, but it is better to use full names for variable naming ((

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #513, opened 2023-06-01
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}"
+                )
```
YuriiKindrat, line 19: bad formatting here(

YuriiKindrat (approved):

app/main.py
YuriiKindrat: fix this

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #514, opened 2023-06-01
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1 +1,43 @@
...
+            f"{{Name: {self.name}, Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+
+        return self.hidden
```
YuriiKindrat, line 30: why do u need return here?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> int:
+        if not herbivore.hidden and isinstance(herbivore, Herbivore):
+            herbivore.health -= 50
+
+            if herbivore.health <= 0:
+                Animal.alive.remove(herbivore)
+
+            return herbivore.health
```
LiudmylaKulinchenko, line 41: No need to return anything in the `bite` function

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 34: Fix annotations for `herbivore` here, please. If it could be only Herbivore, you wouldn't need the next check in the function

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🎈

PR #515, opened 2023-06-01
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -1 +1,37 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.hidden = hidden
+        self.health = health
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 21: try to avoid using linebreaker here, u can use brackets instead

app/main.py
```diff
@@ -1 +1,37 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Union[Herbivore, Carnivore]) -> None:
```
YuriiKindrat, line 33: could u use a parent class for type annotation here?

YuriiKindrat (changes requested):

PR #516, opened 2023-06-01
implemented py-herbivores-and-carnivores

mariiahorbova:


app/main.py
```diff
@@ -1 +1,41 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, *args, **kwargs) -> None:
```
YuriiKindrat, line 7: why do u use *args, **kwargs here?

app/main.py
```diff
@@ -1 +1,41 @@
...
+        else:
+            self.health = 100
+        self.name = name
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YuriiKindrat, line 19: try to avoid using linebreakers, u can use brackets instead

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore | Carnivore) -> None:
```
YuriiKindrat, line 37: Could u use Animal for type annotation here?

app/main.py
```diff
@@ -1 +1,41 @@
...
+        if isinstance(target, Carnivore) or target.hidden:
+            return
```
YuriiKindrat, line 39: no need to use blank return here, u can use backward condition here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #517, opened 2023-06-01
Solution

Kirontiko:


app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore | Carnivore) -> None:
```
YuriiKindrat, line 33: could u use parent class for type annotation here?

YuriiKindrat (changes requested):

app/main.py
Kirontiko: > could u use parent class for type annotation here?

Can't understand. Like this? Herbivore(Animal) | Carnivore(Animal)

YuriiKindrat (approved):

PR #520, opened 2023-06-02
Solution

erikagayan:


app/main.py
```diff
@@ -1 +1,27 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 22: you don't need self here, so make this method static

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌺

PR #521, opened 2023-06-02
'Solution'

RedCatHot:


app/main.py
```diff
@@ -1 +1,27 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
LiudmylaKulinchenko, line 11: U can use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,27 @@
...
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Animal) -> None:
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
```
LiudmylaKulinchenko, line 21: Move this `if` condition to the upper level to avoid nested conditions

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice work 🌺🌼☀

PR #522, opened 2023-06-02
My solution

kovaliskoveronika:


app/main.py
```diff
@@ -1 +1,36 @@
...
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
LiudmylaKulinchenko, line 17: U can use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,36 @@
...
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    @staticmethod
+    def still_alive(animal: Animal) -> None:
+        if animal.health <= 0:
+            animal.alive.pop(animal.alive.index(animal))
```
LiudmylaKulinchenko, line 24: You can remove specified animal from the Animal class `alive` variable using `remove` method (same as in your __init__ method,  just use `remove`)

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌺

PR #523, opened 2023-06-02
Implemented Animal, Herbivore and Carnivore

Altahoma:


app/main.py
```diff
@@ -1 +1,36 @@
...
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    def is_alive(self) -> bool:
+        return self.health > 0
+
+    def take_damage(self, damage: int) -> None:
+        self.health -= damage
+        if not self.is_alive() and self in Animal.alive:
```
LiudmylaKulinchenko, line 22: You can simplify this condition. Remove `is_alive` function and put `self.health > 0` directly, no need to create function for it. And you can skip check "self in Animal.alive"

LiudmylaKulinchenko (changes requested):

app/main.py
Altahoma: Adding "is_alive()" improves readability and future scalability, we can read it almost like usual english text and non-programmer will understad what does it mean.
For tests it's no need checking "self in Animal.alive" but in task we have "Nature is well suited to reflect the principles of Object Oriented Programming.", so if lion try to bite the rabbit's dead body, what usual case, code without this check will produce an error.

LiudmylaKulinchenko: Your function should make some action, so use the boolean variable `is_alive` instead. It is named perfect for the boolean variable, what do you think?

LiudmylaKulinchenko (changes requested):

app/main.py
Altahoma: Function doing execly what's expected, cheking ojcect alive or not. Adding `is_alive` directly into class attribute have a similar result, just another way and would probably also require a method for changing the status if I don't want to change it not in an obvious way. That's why I wrote a `take_damage()` avoiding direct access to variables from another class.

LiudmylaKulinchenko: We require that the task be performed in accordance with all best practices. It is **bad practice** to create a function or method just for not complecated boolean condition & **bad practice** to name your funciton like that. Please, explain me why do you disagree with these 2 statements and I could answer you directly on the question 🍬

LiudmylaKulinchenko (changes requested):
I've added you some comments to the discussion 🐇

app/main.py
Altahoma: "is" is a verb.

> A predicate is a function that returns the truth value of some condition;
`def is_even(x):
    return (x % 2) == 0`
[docs.python.org](https://docs.python.org/3/howto/functional.html)

LiudmylaKulinchenko: Your function could be just a variable, so delete redutant function, please, I will not approve it. It is a bad pratice, you could use less code, we want you to write your code during to the best pratices, so you could be the best candidates on the labor market 😊

LiudmylaKulinchenko: I meant method* but I think you get it 😎

LiudmylaKulinchenko (approved):
Well done! 🌺

PR #524, opened 2023-06-02
Solution

YevheniiIovchu:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 31: Fix annotations for `herbivore` here, plere. If it could be only Herbivore, you wouldn't need the next check in the function

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
:+1:

PR #527, opened 2023-06-02
Solution

Kasatello:


app/main.py
```diff
@@ -1 +1,35 @@
...
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.hidden = hidden
+        self.health = health
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
LiudmylaKulinchenko, line 15: U can use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,35 @@
...
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def remove_from_alive(self) -> None:
```
LiudmylaKulinchenko, line 19: Remove this method

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
+        herbivore.health -= 50
+        if herbivore.health <= 0:
+            herbivore.remove_from_alive()
```
LiudmylaKulinchenko, line 35: Make actions from `remove_from_alive` method here directly

LiudmylaKulinchenko (changes requested):

app/main.py
Kasatello: but then flake8 says dat the line is too long

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
```
YuriiKindrat, line 31: why do u need blank return here? could u use backward condition here?

YuriiKindrat (changes requested):

PR #528, opened 2023-06-04
Solution

Po-SvinSkipodla:
I`m not sure about this linebreaks

```python
def __repr__(self) -> str:
        return f"{{Name: {self.name}, " \
               f"Health: {self.health}, " \
               f"Hidden: {self.hidden}}}"
```

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,32 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
LiudmylaKulinchenko, line 16: Appreciate your attention, you really could use brackets instead of linebreaker here 🌺

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: "Herbivore") -> None:
+        if not isinstance(victim, Carnivore) and not victim.hidden:
```
LiudmylaKulinchenko, line 29: You could just check if it is a `Herbivore` instance to simplify

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🌼🌺

PR #531, opened 2023-06-05
'Solution'

olmatiash:
Hello!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: str = 100, hidden: bool = False)\
+            -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if herbivore.hidden is False:
+                herbivore.health -= 50
+            print(f"{self} cannot bite hidden {herbivore}")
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,31 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if herbivore.hidden is False:
+                herbivore.health -= 50
+            print(f"{self} cannot bite hidden {herbivore}")
```
ArsenPidhoretskyi, line 27: Remove this print. FYI it always print, also when hidden is False

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if herbivore.hidden is False:
```
ArsenPidhoretskyi, line 25: Use and

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    def __init__(self, name: str, health: str = 100, hidden: bool = False)\
+            -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
vasylhnatiuk, line 12: use () instead of \ for long str

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: str = 100, hidden: bool = False)\
+            -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
vasylhnatiuk, line 9:
```suggestion
        self.alive.append(self)
```

vasylhnatiuk (changes requested):

PR #538, opened 2023-06-13
Solution

Veinmax:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if isinstance(other, Herbivore) and other.hidden is False:
+            other.health -= 50
+            Animal.is_alive()
```
ArsenPidhoretskyi, line 39: Maybe better to check just this one animal? Because if we will have 1000000 objects it will be so long to check all objects when only one can be less < 0

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
vsmutok, line 34: image

app/main.py
```diff
@@ -1 +1,42 @@
...
+    @classmethod
+    def is_alive(cls, animal: annotations) -> None:
+        if animal in cls.alive:
+            if animal.health <= 0:
+                cls.alive.remove(animal)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
vsmutok, line 26: better to user `()` then `\`

app/main.py
```diff
@@ -1 +1,42 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    @classmethod
+    def is_alive(cls, animal: annotations) -> None:
+        if animal in cls.alive:
+            if animal.health <= 0:
+                cls.alive.remove(animal)
```
vsmutok, line 21: it seems to me that this is superfluous here

app/main.py
```diff
@@ -1 +1,42 @@
...
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if isinstance(other, Herbivore) and other.hidden is False:
+            other.health -= 50
+            Animal.is_alive(other)
```
vsmutok, line 42:
```suggestion
            Animal.alive.remove(other)
```

vsmutok (changes requested):

PR #540, opened 2023-06-16
'Solution 16/06/23'

TymchenkoV:


app/main.py
```diff
@@ -1 +1,33 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Grayder0152, line 19: use `()` instead of `\`

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
vasylhnatiuk, line 14:
```suggestion
       self.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore_obj: Herbivore | Carnivore) -> None:
+        if isinstance(herbivore_obj, Herbivore) and \
+                herbivore_obj.hidden is False:
+            herbivore_obj.health -= 50
+            if herbivore_obj.health <= 0:
+                Animal.alive.remove(herbivore_obj)
```
vasylhnatiuk, line 33:
```suggestion
               self.alive.remove(herbivore_obj)
```

app/main.py
TymchenkoV: is there some principal difference in these two cases?

app/main.py
```diff
@@ -1 +1,33 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore_obj: Herbivore | Carnivore) -> None:
```
ArsenPidhoretskyi, line 28: Wrong type annotations. It can only be Herbivore
Also rename herbivore_obj to herbivore

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #544, opened 2023-06-20
Implemented Herbivores-and-carnivores task

OleksiiKiva:


app/main.py
```diff
@@ -1 +1,37 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: "Animal") -> None:
```
YukihiroSM, line 32: What about seetting type hint to "Herbivore"? Or it can be any animal? In this case, what about changing variable name to something neutral? Decide what to do here. Change type hint or change variable name :)

YukihiroSM (changes requested):
Overall everything look good, excepting one comment. Good job!

YukihiroSM (approved):
Good job!

PR #548, opened 2023-06-27
'Solution'

govorunenko:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YukihiroSM, line 13: Don't use `\` for line separation. Better use `(` `)`.

app/main.py
```diff
@@ -1 +1,33 @@
...
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> None:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            Animal.checking_in_alive()
```
YukihiroSM, line 33: If you have 1 million animals - you will need to go through this loop every time, when one of it dies. Just imagine the load. 
What about just removing dead animal from list?

YukihiroSM (changes requested):

app/main.py
govorunenko: Done.

app/main.py
govorunenko: I've made some changes. Hope there're correct.

YukihiroSM: The changed code basically is doing the same. You are looping through all animals to delete one of them. 
Why don't you just remove the exact `other` animal from alive list if it's health <= 0?

YukihiroSM (changes requested):

app/main.py
govorunenko: Sorry. Now I've deleted method "checking_in_alive" and expanded functionality of the method "bite". It checks if the animal is dead at once.

app/main.py
```diff
@@ -1 +1,30 @@
...
+            self.hidden = False
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> None:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            if other.health <= 0:
+                Animal.alive.remove(other)
```
YukihiroSM, line 30: This check can be done for whole bite method not only for Herbivore (becuase we are awaiting Animal as `other`). But that's a recomedation for future. No need to fix it now. 
reason: in future if you will add something like biting other Carnivores or something like that - you will need to duplicate the same check to the other if statement.

YukihiroSM (approved):
All looks good now! Thanks for the job! 
Read the comment also, I shared some of my thoughts there, 
Approved!

app/main.py
govorunenko: Understood. Thank you very much!

PR #558, opened 2023-07-04
Solution

anastmishchuk:


app/main.py
```diff
@@ -1 +1,30 @@
...
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
vasylhnatiuk, line 13: use () instead of \

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
ArsenPidhoretskyi, line 10: You also can do:
self.alive.append(self)

ArsenPidhoretskyi (approved):
Good job!

PR #559, opened 2023-07-04
Solution

nschcur:


app/main.py
```diff
@@ -1 +1,47 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
```
vasylhnatiuk, line 19: use () instead of \

app/main.py
```diff
@@ -1 +1,47 @@
-# write your code here
+from typing import List
+
+
+class Animal:
+    alive: List["Animal"] = []
```
```suggestion
    alive: List[Animal] = []
```

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def die(self) -> None:
+        self.health = 0
+        self.hidden = True
+        self.alive.remove(self)
+
+    def bite(self, animal: object) -> None:
```
vasylhnatiuk, line 27:
```suggestion
    def bite(self, animal: object) -> None:
```
```suggestion
    def bite(self, animal: Animal) -> None:
```

app/main.py
vasylhnatiuk: add from future import __annotations__

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #575, opened 2023-07-19
solution

H1kki:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def is_hiding(self) -> bool:
+        return self.hidden
+
+    @staticmethod
+    def remove_dead_animals() -> None:
+        Animal.alive = [animal for animal in Animal.alive if animal.health > 0]
```
Dimosphen1, line 28: There is no need for this `staticmethod`, because list comprehension with iteration would be excessive when you know what exact herbivore should be removed from the `alive` list in the `bite` method.

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden  # Toggle the value of 'hidden'
+        if self.hidden:
+            print(f"{self.name} is now hiding.")
+        else:
+            print(f"{self.name} is no longer hiding.")
```
Dimosphen1, line 37: There is no need for this logic, remove it.

app/main.py
```diff
@@ -1 +1,52 @@
...
+        return self.hidden
+
+    @staticmethod
+    def remove_dead_animals() -> None:
+        Animal.alive = [animal for animal in Animal.alive if animal.health > 0]
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden  # Toggle the value of 'hidden'
```
Dimosphen1, line 33: This operation is descriptive by itself, remove the comment.

app/main.py
```diff
@@ -1 +1,52 @@
...
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}"
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def is_hiding(self) -> bool:
+        return self.hidden
```
Dimosphen1, line 24: You could remove this method and call directly `self.hidden` in places when needed.

app/main.py
```diff
@@ -1 +1,52 @@
...
+            herbivore.health -= 50
+            print(
+                f"{self.name} bit {herbivore.name}. "
+                f"{herbivore.name} health: {herbivore.health}"
+            )
+            if herbivore.health <= 0:
+                print(f"{herbivore.name} is dead.")
+                Animal.remove_dead_animals()
+        else:
+            print(f"{self.name} cannot bite {herbivore.name}.")
```
Dimosphen1, line 52: Remove `print` statements from this method, there is no need for them.

Dimosphen1, line 52: Rewrite this method to not use `else`.

app/main.py
```diff
@@ -1 +1,52 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return f"Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}"
```
Dimosphen1, line 16: Avoid using line breakers when using multiple strings, return a string in `()`. It will allow you to omit using `\`.

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Dimosphen1, line 21: Return string without using line breaks, put it into `()`

Dimosphen1 (changes requested):
Several changes were requested,

Polyakiv-Andrey (approved):
Good job!

PR #578, opened 2023-07-21
Solution

sachavskyi:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
```
Polyakiv-Andrey, line 5: add hidden in init arguments with default value False

app/main.py
```diff
@@ -1 +1,31 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = False
```
Polyakiv-Andrey, line 8: do the same as other attributes

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+                f" Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
pavlejviki, line 27: you can simplify it:

self.hidden = not self.hidden

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+    def __init__(
+        self, name: str,
+        health: int = 100,
+        hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
pavlejviki, line 13: you can use self.alive instead

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #581, opened 2023-07-25
task completed

andriy-py:


app/main.py
```diff
@@ -1 +1,60 @@
...
+    alive: List["Animal"] = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        """
+        Initialize an animal with a name and health.
+        """
+        self.name = name
+        self.health = health
+        self.hidden = False
+        self.__class__.alive.append(self)
```
ArsenPidhoretskyi, line 17:
```suggestion
        self.alive.append(self)
```
`__class__` is redundant

ArsenPidhoretskyi: Fix in other places

ArsenPidhoretskyi (changes requested):

app/main.py
ArsenPidhoretskyi: not fixed

ArsenPidhoretskyi (changes requested):

vsmutok (changes requested):

PR #587, opened 2023-07-27
'Solution'

niksgawrilenko:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
```
Dimosphen1, line 9: It would be better to place `self` on a new line here so that `(` and `)` will not have[ any arguments along with them](https://google.github.io/styleguide/pyguide.html#3192-line-breaking), e.g.:
 ```suggestion
    def __init__(
            self,
            name: str,
            health: int = 100,
            hidden: bool = False
    ) -> None:
```

Dimosphen1 (approved):
LGTM, one minor comment is left for review.

PR #591, opened 2023-08-02
Classes that inherit from Animal class.

IProskurnytskyi:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
```
Dimosphen1, line 18: There is no need for these variables, call `self.some_value` directly in the f string.

Dimosphen1 (changes requested):
Several minor changes were requested.

PR #594, opened 2023-08-08
Solution

nyanplague:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
```
Y-Havryliv, line 4: hidden is false by default, but if I want to create an animal with hidden=True I will get an error.

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
```
Y-Havryliv, line 8: please read the checklist

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
```
Y-Havryliv, line 11: using `()` would be more readable

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if herbivore.hidden or not isinstance(herbivore, Herbivore):
```
Y-Havryliv, line 24: it's better to do this check:
`if isinstance(herbivore, Herbivore) and not herbivore.hidden:`

Y-Havryliv (changes requested):

app/main.py
Kroll410: Animal.alive instead of self.alive, because self - reference for the instance of the class, Animal - class by itself

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+        return (
+            f"{{Name: {self.name},"
+            f" Health: {self.health},"
+            f" Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
```
Kroll410, line 22: Use = False instead of none, because `hide` method shouldn't unhide an animal

Kroll410 (changes requested):

IvanRamyk (approved):

PR #596, opened 2023-08-09
Solution

eduardhabryd:


app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+    def remove_from_alive(self) -> None:
+        Animal.alive.remove(self)
```
ORENYT, line 19: I think mentor will reject this code, because you don't need this method inside class. You call this only one time.

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.health = 0
```
ORENYT, line 35: I've runned tests without `herbivore.health = 0` , all 12 passed. I think this string is redundant.

app/main.py
```diff
@@ -1 +1,36 @@
...
+                herbivore.remove_from_alive()
```
ORENYT, line 36: You can change this string into single function call `Animal.alive.remove(herbivore)`

ORENYT (commented):
I think mentor will reject this code

app/main.py
ORENYT: But I don't know if it is a best practice to call class element like this. We didn't hear about this. So just wait a mentor's review.

app/main.py
eduardhabryd: > I think mentor will reject this code, because you don't need this method inside class. You call this only one time.

I think there is an OOP principle that states the following: 

Encapsulation is one of the fundamental concepts in Object-Oriented Programming (OOP) that involves bundling data (attributes) and methods (functions) that operate on that data into a single unit, known as a class. 

That's why this method, in my opinion, should be inside the class.

app/main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
```
Kroll410, line 4: Use [this](https://stackoverflow.com/a/41591509) styling for parameters with typings

app/main.py
```diff
@@ -1 +1,35 @@
...
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
Kroll410, line 14: You can simply use + between string lines

app/main.py
Kroll410: That's all right, because Animal is an abstraction that contains the alive ancestors

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            "{"
+            + f"Name: {self.name}, "
+            + f"Health: {self.health}, "
+            + f"Hidden: {self.hidden}"
+            + "}"
+        )
```
Y-Havryliv, line 22:
```suggestion
        return (f"{{Name: {self.name}, "
                f"Health: {self.health}, "
                f"Hidden: {self.hidden}}}")
```

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #597, opened 2023-08-09
'Solution'

lgSilay:


app/main.py
```diff
@@ -1 +1,29 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if herbivore.hidden or not isinstance(herbivore, Herbivore):
+            pass
+        else:
```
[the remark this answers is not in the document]
lgSilay: Thank you for your comment, i'll correct this!)

app/main.py
```diff
@@ -1 +1,27 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
```
Kroll410, line 11: In modern stylings people try to avoid `\` enclosure, so you can simply use '+' here

Kroll410 (changes requested):
Looks good

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
```
Y-Havryliv, line 4: hidden is false by default, but if I want to create an animal with hidden=True I will get an error.

app/main.py
```diff
@@ -1 +1,27 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
```
Y-Havryliv, line 8: please read the checklist

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
```
Y-Havryliv, line 13: still not fixed

Y-Havryliv (changes requested):

PR #598, opened 2023-08-10
implemented solution

innyshka:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
```
Kroll410, line 4: Use [this ](https://stackoverflow.com/a/41591509)formatting for functions

app/main.py
```diff
@@ -1 +1,31 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if (animal.hidden is False) and isinstance(animal, Herbivore):
```
Kroll410, line 28: `is` keyword shouldn't be used with booleans, only with None or pointer (memory address) comparing

Kroll410 (changes requested):
Looks good

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if (not animal.hidden) and isinstance(animal, Herbivore):
```
Y-Havryliv, line 31:
```suggestion
        if not animal.hidden and isinstance(animal, Herbivore):
```
you can just delete parentheses

Y-Havryliv (approved):
GJ!

PR #600, opened 2023-08-10
Solution

denys-source:


app/main.py
```diff
@@ -1 +1,36 @@
...
+class Animal:
+
+    alive: list[Animal] = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name: str = name
+        self.health: int = max(min(health, 100), 0)
```
Y-Havryliv, line 13: based on the task I assume that 100 is not the max health, and it's hard to read, so it's better not to complicate things

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Herbivore) -> None:
```
Y-Havryliv, line 32: seems like it's static method

app/main.py
```diff
@@ -1 +1,36 @@
...
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health = max(animal.health - 50, 0)
```
Y-Havryliv, line 34: again it's hard to understand the code at first glance, just subtract 50 and check if health <=0

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
```
Kroll410, line 26: Use better = True, because hide shouldn't unhide animal

Kroll410 (changes requested):

app/main.py
denys-source: I agree, but the tests explicitly check if this method changes value to the opposite

PR #601, opened 2023-08-10
Solved Task. Created 2 classes and __repr__ to represent class as a s…

ORENYT:
Created 2 classes and __repr__ to represent class as a string (fake dict).

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,34 @@
-# write your code here
+from typing import Callable
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(bitten_animal: "Animal") -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,36 @@
...
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
```
Y-Havryliv, line 14: using `()` would be more readable

app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(bitten_animal: "Animal") -> None:
+        if isinstance(bitten_animal, Carnivore):
```
Y-Havryliv, line 28: to be honest, it's hard to read a bit, you can just write:
`if isinstance(herbivore, Herbivore) and not herbivore.hidden:`
and you can delete animal from the list using `remove`

Y-Havryliv (changes requested):

app/main.py
ORENYT: Changed self to Animal.

app/main.py
ORENYT: Fixed this.

app/main.py
ORENYT: used this

PR #605, opened 2023-08-10
Solution

katryana:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
```
Y-Havryliv, line 4: use [this](https://stackoverflow.com/questions/24988162/define-functions-with-too-many-arguments-to-abide-by-pep8-standard/41591509#41591509) formatting

app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore_animal: Herbivore) -> None:
+        check_class = isinstance(herbivore_animal, Herbivore)
+        check_hidden_value = herbivore_animal.hidden
+        if check_class and not check_hidden_value:
```
Y-Havryliv, line 33: you can create additional variables for readability but I suggest better naming
`if is_herbivore_instance and not is_hidden:`
they tell us more info

Y-Havryliv (changes requested):

Kroll410 (approved):

PR #608, opened 2023-08-11

Barsh4ec:


app/main.py
```diff
@@ -1 +1,33 @@
...
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
Y-Havryliv, line 10: hidden is false by default, but if I want to create an animal with hidden=True I will get an error.

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+    def __repr__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".format(
```
Y-Havryliv, line 14: `f-strings` are preferable, they are way more readable

app/main.py
```diff
@@ -1 +1,33 @@
...
+            self.name,
+            self.health,
+            self.hidden)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
```
Y-Havryliv, line 22: please read the checklist

app/main.py
```diff
@@ -1 +1,33 @@
...
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
Y-Havryliv, line 29: seems like it's static method

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+from __future__ import annotations
```

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: still no fixed

app/main.py
```diff
@@ -1 +1,31 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
Y-Havryliv, line 16: using `()` would be more readable

app/main.py
```diff
@@ -1 +1,31 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not herbivore.hidden and isinstance(herbivore, Herbivore):
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                super().alive.remove(herbivore)
```
Y-Havryliv, line 31: just use class name

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ! go get some rest))

PR #610, opened 2023-08-11
Implemented animal classes and it's methods

MarianKovalyshyn:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+from __future__ import annotations
```

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
```
Y-Havryliv, line 7: hidden is false by default, but if I want to create an animal with hidden=True I will get an error.

Y-Havryliv (changes requested):

app/main.py
MarianKovalyshyn: Idk, maybe it was needed before my final code and I forgot about this.

Y-Havryliv (approved):
Nice!

PR #612, opened 2023-08-11
Solution

OleksandrMoroz24:


app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from __future__ import annotations
```

app/main.py
```diff
@@ -1 +1,32 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
+
+    def __str__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".format(
```
Y-Havryliv, line 14: `f-strings` are preferable, they are way more readable

app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
```
Y-Havryliv, line 7: hidden is false by default, but if I want to create an animal with hidden=True I will get an error.

app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = False
+        self.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,32 @@
...
+        return self.__str__()
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Herbivore) -> None:
```
Y-Havryliv, line 28: seems like it's static method

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
+                super().alive.remove(target)
```
Y-Havryliv, line 32: just use class name

Y-Havryliv (changes requested):

app/main.py
Kroll410: Use Animal.alive instead of self.alive

Kroll410 (changes requested):

Kroll410 (approved):

PR #616, opened 2023-08-11
Solution

ochernous:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = False if self.hidden else True
```
Y-Havryliv, line 23:
```suggestion
        self.hidden = not self.hidden
```

app/main.py
```diff
@@ -1 +1,31 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
Y-Havryliv, line 16: use `()`, it's more readable

app/main.py
```diff
@@ -1 +1,31 @@
...
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = False if self.hidden else True
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore_obj: Herbivore) -> None:
```
Y-Havryliv, line 27: seems like it's static method

Y-Havryliv, line 27: I'm strongly against names like this herbivore_obj. You can just use herbivore

Y-Havryliv (changes requested):

app/main.py
ochernous: Ok) Can I ask you, why? What problems can naming like herbivore_obj cause?

app/main.py
```diff
@@ -1 +1,34 @@
...
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
```
Kroll410, line 25: Pass self.hidden = False, because hide method shouldn't unhide animal

Kroll410 (changes requested):

app/main.py
ochernous: But here we see that hide method can unhide animal [https://imgur.com/a/cay6hVG](url)

Kroll410: Got it, bad description. Approved, good job

Kroll410 (approved):

PR #618, opened 2023-08-12
hac_v1

VASO304:


app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self,
```
Kroll410, line 5: Formatting!

app/main.py
```diff
@@ -1 +1,33 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if (animal.hidden is False) and isinstance(animal, Herbivore):
```
Kroll410, line 30: use `==` instead of `is`. Check [this](https://towardsdatascience.com/whats-the-difference-between-is-and-in-python-dc26406c85ad)

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(
```
Y-Havryliv, line 5: better to format like [this](https://stackoverflow.com/a/41591509/21402604)

Y-Havryliv (changes requested):

IvanRamyk (approved):

PR #623, opened 2023-08-13
herbivores-and-carnivores

stamaksim:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
```
vsmutok, line 8:
```suggestion
        self.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __repr__(self) -> str:
+        return f"{{" \
+            f"Name: {self.name}," \
+            f" Health: {self.health}," \
+            f" Hidden: {self.hidden}" \
+            f"}}"
```
vsmutok, line 15: better to use `()` then `\`

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
vsmutok, line 23: image

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, other: str) -> int:
```
vsmutok, line 27:
```suggestion
    def bite(self, other: Herbivore) -> None:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+        if isinstance(other, Carnivore):
+            self.health = 100
```
vsmutok, line 29: you must check is other is Herbivore

vsmutok (changes requested):

app/main.py
vsmutok: do it

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Herbivore) -> None:
+        if isinstance(other, Carnivore):
+            self.health = 100
```
vsmutok, line 28: Redundant here

vsmutok (changes requested):

Grayder0152 (approved):

PR #629, opened 2023-08-17
Solution py-herbivores-and-carnivores

IvDyriv:


app/main.py
```diff
@@ -1 +1,31 @@
...
+                 ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
YukihiroSM, line 17: Don't use `\` for line separation. Use `(` `)` instead.

YukihiroSM (changes requested):

app/main.py
IvDyriv: Okay, thank you :)

YukihiroSM (approved):
looks good to me!

PR #641, opened 2023-08-24
Solution

Lebwa1337:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
Dimosphen1, line 18: Return the f-string in `()` so that you won't need `\` for line breaking in the return statement.

app/main.py
```diff
@@ -1 +1,33 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal_to_bite: Herbivore) -> None:
+        is_it_carnivore = not isinstance(animal_to_bite, Carnivore)
+        if animal_to_bite.hidden is False and is_it_carnivore:
```
Dimosphen1, line 30: There is no need for the `is_it_carnivore` variable. Add it to the if condition directly. Besides that, evaluating the `.hidden` attribute with `is` is over-explicit here, just assume that it will always have a boolean value and build this if condition using `not`.

Dimosphen1 (changes requested):

app/main.py
Lebwa1337: if i don`t use is_it_carnivore variable,then flake8 will that "line is to long" . Same with return f-string in 16 line

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal_to_bite: Herbivore) -> None:
+        is_it_carnivore = not isinstance(animal_to_bite, Carnivore)
```
ArsenPidhoretskyi, line 29: Why you check for Carnivore?
If animal_to_bite must be Herbivore

ArsenPidhoretskyi (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #647, opened 2023-08-28
Herbivore and Carnivore Solution

Filllick:


app/main.py
```diff
@@ -1 +1,31 @@
...
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
vsmutok, line 15: better to use `()` then `\`

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
vsmutok, line 23: no need to use `if`, use only `not` 

image

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not herbivore.hidden and isinstance(herbivore, Herbivore):
+            herbivore.health -= 50
+        if herbivore.health <= 0:
+            super().alive.remove(herbivore)
```
vsmutok, line 31:
```suggestion
           self.alive.remove(herbivore)
```

vsmutok (changes requested):

Filllick:
in __repr__ method added parentheses for better breaks, changed hide and bite methods

vsmutok (approved):
Well Done 👍

PR #658, opened 2023-09-04
Solved

Dmitriy-Poplinski:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, beast: Animal) -> None:
+        if not beast.hidden and isinstance(beast, Herbivore):
+            beast.health = beast.health - 50
+            if beast.health <= 0:
+                super().alive.remove(beast)
```
Dimosphen1, line 37: You should call alive from class `Animal` because it is a class attribute, remove `super()` call here.

app/main.py
```diff
@@ -1 +1,37 @@
...
+        ])
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, beast: Animal) -> None:
```
Dimosphen1, line 33: The only animal that can be bitten is `Herbivore`, change your type annotation to encounter that.

app/main.py
```diff
@@ -1 +1,37 @@
...
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return ", ".join([
```
Dimosphen1, line 20: Write `__repr__` without using `join`, there is no benefit for declaring a list and iterating through it here. Write the f-string in a whole inside `()` here to avoid line breaks.

app/main.py
```diff
@@ -1 +1,37 @@
...
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        self.alive.append(self)
```
Dimosphen1, line 17: Call `alive` not from `self`, but rather from the `Animal` class, it is a class attribute.

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Animal:
+    alive: list = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
```
Dimosphen1, line 11: `health` and `hidden` parameters are optional, encounter that in the type annotation.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #668, opened 2023-09-14
Solution

weuis:


app/main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int | None = 100,
+            hidden: bool | None = False
```
viktoria-rybenchuk, line 8: type annotations is incorrect because it suggests that these parameters can be assigned the value None, which contradicts their intended use of having default values 100 and False, respectively

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #672, opened 2023-09-15
'Solution'

nsdNite:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Animal) -> None:
```
Dimosphen1, line 30: You should indicate here that `other` animal that is possible to bite should be a `Herbivore` instance, change this in the type annotation.

Dimosphen1 (changes requested):
One minor change was requested to a type annotation.

app/main.py
nsdNite: done, thanks for the comment

Dimosphen1 (approved):
LGTM, well done!

PR #674, opened 2023-09-18
'Solution'

tiron-vadym:


app/main.py
```diff
@@ -1 +1,32 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def die(self) -> None:
+        Animal.alive.remove(self)
+
+    def __str__(self) -> str:
```
Dimosphen1, line 18: You should redefine `__repr__` method here, not `__str__`.

app/main.py
```diff
@@ -1 +1,32 @@
...
+        return f"{Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}}"
```
Dimosphen1, line 19: Your string formatting is not correct, it should be 
```suggestion
        return f"{{Name: {self.name}, Health: {self.health}, Hidden: {self.hidden}}}"
```
You could add line breaking inside of `()` to improve the readability,

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore) -> None:
```
Dimosphen1, line 28: You should add type annotation for the `herbivore` variable here.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

app/main.py
tiron-vadym: Why should I use here double "{"?

PR #677, opened 2023-09-19
Develop

asdadaversa:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore_object: Herbivore) -> None:
+        if isinstance(herbivore_object, Carnivore) or herbivore_object.hidden:
+            pass
```
pavlejviki, line 32: you can remove this if statement after implementing the suggestion under this comment

pavlejviki (changes requested):

asdadaversa:
its a final

app/main.py
```diff
@@ -1 +1,35 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore_object: Herbivore) -> None:
```
Dimosphen1, line 30: There is no need to name this variable as `herbivore_object` because everything in Python is an object. Remove the redundant `_object` part from the variable naming.

app/main.py
```diff
@@ -1 +1,35 @@
...
+        if isinstance(herbivore_object, Herbivore) and \
```
Dimosphen1, line 31: There would be no need for line breaking after shortening the variable name, besides that, you could shorten the last part with `and not herbivore.hidden:`.

app/main.py
```diff
@@ -1 +1,35 @@
...
+            hidden: bool = False
+    ) -> None:
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
```
Dimosphen1, line 17: Do not apply line-breaking here for string, return string within `()`. It will allow omit using the `\` operator.

Dimosphen1 (changes requested):

app/main.py
asdadaversa: don't understand this comment

asdadaversa:
done, but don't understand why this way is better

viktoria-rybenchuk:
> done, but don't understand why this way is better

viktoria-rybenchuk (approved):
Good job!

PR #679, opened 2023-09-24
Solution

PyCore23:


app/main.py
```diff
@@ -1 +1,38 @@
...
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Carnivore(Animal):
+
+    def bite(self, target: Union) -> None:
+        target.health -= 50 \
+            if (isinstance(target, Herbivore) and not target.hidden) else 0
```
Dimosphen1, line 31: Write this code in multiple lines to avoid line-breaking with `\`. Besides that, if the `target` is an instance of `Carnivore`, or it is hidden its health would always equal `0` in your current version.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Carnivore(Animal):
+
+    def bite(self, target: Union) -> None:
```
Dimosphen1, line 29: Change your type annotation for the target to indicate that you expect a `Herbivore` instance.

app/main.py
```diff
@@ -1 +1,38 @@
...
+        target.health -= 50 \
+            if (isinstance(target, Herbivore) and not target.hidden) else 0
+        Animal.remove_dead_animal()
```
Dimosphen1, line 32: It would be inefficient to make a perform list comprehension and create a new list each time after the animal is deleted from the `alive`. Use `.remove(target)` to perform deletion in place and implement this logic inside of the `bite` method, delete the `remove_dead_animal` class method from the `Animal` class.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
Dimosphen1, line 15: Refer to `Animal.alive` instead of `self.alive`, because `alive` is a class attribute.

app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from typing import Union
```
Dimosphen1, line 1: Do not forget to remove unnecessary imports.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False) -> None:
```
Dimosphen1, line 11: Place `) -> None:` on a new line to format your code properly.

Dimosphen1 (changes requested):

app/main.py
pavlejviki: not fixed

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
viktoria-rybenchuk, line 17: Append the current instance to the class attribute 'alive'

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    @classmethod
+    def remove_dead_animal(cls, animal: Type[Herbivore]) -> None:
```
viktoria-rybenchuk, line 20:
```suggestion
    def remove_dead_animal(cls, animal: Herbivore) -> None:
```

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Well done!

PR #682, opened 2023-09-26
solution: herbivores and carnivores

sacrosanctry:


app/main.py
```diff
@@ -1 +1,42 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and herbivore.hidden is False:
```
viktoria-rybenchuk, line 35: avoid unnecessary 'is False' comparison

app/main.py
```diff
@@ -1 +1,42 @@
...
+            herbivore.health -= 50
+
+            if herbivore.health <= 0:
+                for index, animal in enumerate(Animal.alive):
+                    if animal is herbivore:
+                        Animal.alive.pop(index)
+                        break
```
viktoria-rybenchuk, line 42: why do you need this? You can just remove animal without this loop

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Good job!

PR #686, opened 2023-09-29
feat: Solution

sashokpilipok7:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+    default_health = 100
```
ArsenPidhoretskyi, line 3: This is redundant. Instead use default for health not as None, use 100

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __init__(self, name: str, health: int = None) -> None:
```
ArsenPidhoretskyi, line 5:
```suggestion
    def __init__(self, name: str, health: int = 100) -> None:
```

app/main.py
```diff
@@ -1 +1,31 @@
...
+        return (
+            f"{{Name: {self.name}, Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Carnivore(Animal):
+    def check_is_alive(self, animal: Animal) -> None:
+        if animal.health <= 0:
+            super().alive.remove(animal)
```
ArsenPidhoretskyi, line 21:
```suggestion
           self.alive.remove(animal)
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+        return (
+            f"{{Name: {self.name}, Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Carnivore(Animal):
+    def check_is_alive(self, animal: Animal) -> None:
+        if animal.health <= 0:
+            self.alive.remove(animal)
```
ArsenPidhoretskyi, line 20: Better to move this to Animal class

ArsenPidhoretskyi: And this method need to check that instance alive or not(not parameter)

ArsenPidhoretskyi (changes requested):

app/main.py
sashokpilipok7: @ArsenPidhoretskyi Do you mean that i also have to return true or false?

sashokpilipok7: Oh i'm sorry, i accidentally requested review, i just wanted to ask question above
@ArsenPidhoretskyi

ArsenPidhoretskyi: No you need to move this method to class Animal
It must accept only self, not animal because it is so strange

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
```
vasylhnatiuk, line 8:
```suggestion
       self.alive.append(self)
```

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #688, opened 2023-10-02
Solition

Artem-user-19:


app/main.py
```diff
@@ -1 +1,29 @@
...
+        Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: str) -> None:
```
viktoria-rybenchuk, line 25: use correct type annotations

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+        Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: str) -> list:
```
viktoria-rybenchuk, line 25:
```suggestion
    def bite(self, herbivore:  Herbivore) -> list:
```

viktoria-rybenchuk (changes requested):

app/main.py
Artem-user-19: Why, None in my opinion is better here, I don`t know which type do I must to put here

PR #695, opened 2023-10-05
Solution

Naz-iv:


D1st3f (approved):
All looks good

app/main.py
```diff
@@ -1 +1,43 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health -= 50
+            if animal.health <= 0:
```
LiudmylaKulinchenko, line 42: You don't need this check here, you have one in remove_dead_animal method

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: I added this check to avoid calling remove_dead_animal if animal health is more then 0, but will remove it

fsocie7y (approved):
well done ^_^

PR #699, opened 2023-10-07
Solution

skoval7:


app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> None:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    def __repr__(self) -> None:
+        return str(self)
```
petrykivd, line 21: You can use only __repr__

petrykivd (changes requested):

petrykivd (approved):
GJ🚀

app/main.py
```diff
@@ -1 +1,33 @@
...
+                f" Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None\
```
LiudmylaKulinchenko, line 27: Don't break a line here

app/main.py
```diff
@@ -1 +1,33 @@
...
+            :
+        if isinstance(animal, Carnivore) or animal.hidden:
```
LiudmylaKulinchenko, line 29: Use backward condition here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+                f" Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
```
fsocie7y, line 27: you don't need self here, so make this method static

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #704, opened 2023-10-07
Animals Battle Royal

SelFisher1488:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore) -> None:
+        if not isinstance(target, Carnivore) and not target.hidden:
+            if target.health > 50:
+                target.health -= 50
+            elif target.health <= 50:
+                target.health = 0
+                Animal.alive.pop(Animal.alive.index(target))
```
sakhaline, line 35: You can use just one "remove" method, that allows you to delete object from the list by itself, instead of using pop and index methods.

app/main.py
```diff
@@ -1 +1,35 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore) -> None:
+        if not isinstance(target, Carnivore) and not target.hidden:
+            if target.health > 50:
```
sakhaline, line 31: I think it would be better to simply subtract 50 from "target.health" attribute, then check: if this attribute is less than or equal to zero and if so, delete it. Then your method will have only two conditions (it is not necessary to use nested conditions) so the code will run faster and look cleaner.

app/main.py
```diff
@@ -1 +1,35 @@
...
+                target.health -= 50
+            elif target.health <= 50:
```
mykolamateichuk, line 33: You can avoid using elif just with checking if health is nagative after bite.

mykolamateichuk (changes requested):

sakhaline (approved):

app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
fsocie7y, line 19: U can use brackets instead of linebreaker here

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore) -> None:
+        if not isinstance(target, Carnivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
+                Animal.alive.pop(Animal.alive.index(target))
```
fsocie7y, line 33: better use Animal.alive.remove()

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Herbivore) -> None:
```
fsocie7y, line 29: fix type annotation for target
imho better annotation here is Animal

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Great! 🌺

PR #707, opened 2023-10-08
Implemented necessary classes and methods, but I couldn't find a way to make line 17 shorter without changing the output

raychw:
Implemented necessary classes and methods, but I couldn't find a way to make line 17 shorter without changing the output

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+
+        self.health = health
+        self.name = name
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
+    def bite(cls, herbivore: Animal) -> None:
```
Dimosphen1, line 28: You could explicitly indicate in the type annotation as well that you expect a `Herbivore` instance for the `herbivore` variable, not a general `Animal` class instance.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #708, opened 2023-10-08
Solution

sakhaline:


OlehShumov (approved):

app/main.py
```diff
@@ -1 +1,35 @@
...
+            __class__.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, victim: "Herbivore") -> None:
```
fsocie7y, line 32: you don't need self here, so make this method static

fsocie7y: and dont use quotes in type annotations

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #711, opened 2023-10-08
Solution

OlehShumov:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Animal) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
```
RVChornyy, line 29: return is redundant here. Try to refactor this method in a way not to use "return" and "else"

RVChornyy (changes requested):

fsocie7y (changes requested):

fsocie7y (commented):
well done ^_^

fsocie7y (approved):

PR #714, opened 2023-10-09
Solution

alex329657:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if (animal.hidden or animal not in animal.alive
+                or not isinstance(animal, Herbivore)):
+            return
+        animal.health -= 50
+        if animal.health <= 0:
+            for item in Animal.alive:
+                if item.name == animal.name:
```
ArsenPidhoretskyi, line 29: Really is it possible that other hp can change?

ArsenPidhoretskyi (changes requested):

app/main.py
alex329657: do I need to check in some other way? the list always has 1 predator and 1 herbivore, can there be more?

tests/test_main.py
```diff
@@ -15,7 +15,7 @@ def test_animal_class():
 
 
 def test_animal_constructor():
-    lion = Animal("Lion King")
+    lion = Animal("Lion King", )
```
Yurnerroo, line 18: don't change this file

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if (animal.hidden or animal not in animal.alive
+                or not isinstance(animal, Herbivore)):
+            return
+        animal.health -= 50
+        if animal.health <= 0:
+            Animal.alive.remove(animal)
```
Yurnerroo, line 28: simplify this and get rid of return

Yurnerroo (changes requested):

tests/test_main.py
alex329657: I don't know how I did it)

app/main.py
alex329657: I changed it, but I'm not sure that it's better

Yurnerroo (approved):
Good job!

PR #715, opened 2023-10-09
Solution

roffi37:


RabenkoYevhenii (approved):
It`s ok. You can try also create method smth like "taking_damage" for the Animal, and just call it in the end of your staticmethod instead of using nested check

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    def dead(self) -> None:
+        if self.health <= 0:
+            self.alive.remove(self)
```
fsocie7y, line 19: u should to use class name instead of self)

fsocie7y (changes requested):

app/main.py
roffi37: ok, but could you explain why?)

fsocie7y: `couse alive its class atribute but not object of class)

fsocie7y (changes requested):

PR #716, opened 2023-10-09
Solution

Wonsky1:


roffi37 (approved):

app/main.py
```diff
@@ -1 +1,27 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
fsocie7y, line 13: U can use brackets instead of linebreaker here

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
```
LiudmylaKulinchenko, line 25: use more descriptive name instead of `other`

LiudmylaKulinchenko, line 25: Fix annotations. If your function takes only Herbivore instances you won't need the check in the next line

LiudmylaKulinchenko (changes requested):

app/main.py
Wonsky1: @LiudmylaKulinchenko .Thanks, i updated my code, but it is not a strict annotation, isnt it more understandable to write Herbivore? For example if other programmer will look at that method, he could think that any animal could be bited

fsocie7y (approved):
well done ^_^

PR #719, opened 2023-10-09
Solution

DenPrislipskyi:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    def remove_animal(self) -> None:
```
fedorov-s-od, line 19: Perhaps this function is redundant. It's possible to remove Animal instance in the "bite" method without calling an additional method

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self in Animal.alive:
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is False:
```
fedorov-s-od, line 27: You can rewrite this method in one line )

app/main.py
```diff
@@ -1 +1,39 @@
...
+        )
+
+    def remove_animal(self) -> None:
+        if self in Animal.alive:
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
```
fedorov-s-od, line 26: You can rewrite this method in one line )

fedorov-s-od (approved):

app/main.py
fsocie7y: yeah, fix that)

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 28: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
```
fsocie7y, line 4: fix indentation)

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
```
fsocie7y, line 8: image
u should use one of this variations indentations for cases like this
dont mix they))

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #721, opened 2023-10-09
Solution

MiskoRuslan:


Oxbay (approved):
Interesting solution!

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) is False:
+            return
+        if herbivore.hidden is True or herbivore.health < 1:
+            return
```
fsocie7y, line 33: u can solve this task with one if statement and without return

app/main.py
```diff
@@ -1 +1,35 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 29: you don't need self here, so make this method static

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #722, opened 2023-10-09
'Solution'

fedorov-s-od:


petrykivd (approved):

petrykivd:
All right👍

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 28: you don't need self here, so make this method static

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 29: Fix annotations and parameter name. If this function gets only Herbivore instances, you won't need the check in the 30 line

app/main.py
```diff
@@ -1 +1,34 @@
...
+        if isinstance(herbivore, Herbivore) and herbivore.hidden is not True:
```
LiudmylaKulinchenko, line 30: `herbivore.hidden is not True` looks really tough to understand 😬 Use `not herbivore.hidden`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".format(
```
fsocie7y, line 16: better use f-string instead of .format()

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Great! 🌺

PR #723, opened 2023-10-09
Solution

Andrii-Silenko:


sakhaline (approved):

app/main.py
```diff
@@ -1 +1,26 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{" + f"Name: {self.name}, " \
+                     f"Health: {self.health}, " \
+                     f"Hidden: {self.hidden}" + "}"
```
fsocie7y, line 13: U can use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 22: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,26 @@
...
+        if isinstance(herbivore, Herbivore):
+            herbivore.health -= 50 if not herbivore.hidden else 0
```
fsocie7y, line 24: u can check it in one condition)

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return ("{" + f"Name: {self.name}, "
+                      f"Health: {self.health}, "
+                      f"Hidden: {self.hidden}" + "}")
```
LiudmylaKulinchenko, line 13: U don't need concatenation here, you can just add curly braces directly inside a string

LiudmylaKulinchenko: fix that

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 23: Fix annotation and variable name. If there only Herbivores will be passed in the function, you won't need the check in the next line

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #724, opened 2023-10-09
Solution Herbivores and carnivores

Oxbay:


app/main.py
```diff
@@ -1 +1,36 @@
...
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
fsocie7y, line 28: we can do it in one line)

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, herb: "Herbivore") -> None:
+        if isinstance(herb, Herbivore) and herb.hidden is False:
+            herb.health -= 50
+        if herb.health <= 0:
+            Animal.alive.pop(Animal.alive.index(herb))
```
fsocie7y, line 36: u can use .remove()

app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    def bite(self, herb: "Herbivore") -> None:
```
fsocie7y, line 32: you don't need self here, so make this method static

fsocie7y (changes requested):

app/main.py
Oxbay: yep)

app/main.py
fsocie7y: solve this one too

app/main.py
```diff
@@ -1 +1,33 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herb: "Herbivore") -> None:
```
fsocie7y, line 29: and use annotations without quotes)

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #726, opened 2023-10-10
Solution

Katherine-Greg:


Lyutillis (approved):

app/main.py
```diff
@@ -1 +1,30 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 26: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,30 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        if health > 0:
+            Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{{Name: {}, Health: {}, Hidden: {}}}".format(
```
fsocie7y, line 15: better to use f-string instead of "".format()

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
```
LiudmylaKulinchenko, line 30: Fix annotations and parameter name. If this function gets only Herbivore instances, you won't need the check in the 30 line

LiudmylaKulinchenko (changes requested):

app/main.py
Katherine-Greg: But we need this check, because function can get Carnivore instanсe, and this check is also present in the tests

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
```
LiudmylaKulinchenko, line 29: I answered your comment, fix annotations

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 😸

PR #728, opened 2023-10-11
Carnivores and Herbivores Solution

Lyutillis:


tests/test_main.py
```diff
@@ -129,6 +129,8 @@ def test_print_animal_alive():
 
     f = io.StringIO()
 
+    print(Animal.alive)
+
```
fsocie7y, line 133: dont change the tests))

app/main.py
```diff
@@ -1 +1,31 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if herbivore.hidden or isinstance(herbivore, Carnivore):
+            return
+        herbivore.health = max(0, herbivore.health - 50)
+        if not herbivore.health:
+            Animal.alive.pop(Animal.alive.index(herbivore))
```
fsocie7y, line 31: can we resolve it a little bit simpler?

app/main.py
```diff
@@ -1 +1,31 @@
...
+        return "{" + name + health + hidden + "}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
fsocie7y, line 26: you don't need self here, so make this method static

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = f"Name: {self.name}, "
+        health = f"Health: {self.health}, "
+        hidden = f"Hidden: {self.hidden}"
+        return "{" + name + health + hidden + "}"
```
LiudmylaKulinchenko, line 17: rewrite with only return, use f-strings

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 27: Fix annotations and variable name, if this method would take only Herbivore instances, you won't need a check in the next line

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if herbivore.hidden or isinstance(herbivore, Carnivore):
+            return
```
LiudmylaKulinchenko, line 29: you can use the opposite condition

app/main.py
```diff
@@ -1 +1,32 @@
...
+        herbivore.health -= 50
+        if herbivore.health <= 0:
+            Animal.alive.pop(Animal.alive.index(herbivore))
```
LiudmylaKulinchenko, line 32: u can use `.remove(value)`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+                 hidden: bool = False) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
LiudmylaKulinchenko, line 16: Use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Animal) -> None:
+        if not (herbivore.hidden or isinstance(herbivore, Carnivore)):
```
LiudmylaKulinchenko, line 27: What if later there will be another class that inherits from Animal? You should fix this condition, so it will be True only with Herbivore instance

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Animal) -> None:
```
LiudmylaKulinchenko, line 26: Fix variable name, it is not descriptive now, you use it not only for herbivores

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore | Carnivore) -> None:
```
LiudmylaKulinchenko, line 29: Use Animal for annotations here and fix parameter name, it is not descriptive now

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #732, opened 2023-10-13
Solution

OleksandrMV:


app/main.py
```diff
@@ -1 +1,36 @@
...
+            other.health -= 50
+            Animal.death(other)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
```
EdAlekseiev, line 36: You can simplify this code

EdAlekseiev, line 36:
```suggestion
        self.hidden = not self.hidden
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+                f"Hidden: {self.hidden}}}")
+
+    def death(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if not other.hidden and not isinstance(other, Carnivore):
```
ArsenPidhoretskyi, line 26:
```suggestion
        if not other.hidden and isinstance(other, Herbivore):
```
Due to task we can only bite Herbivore

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #736, opened 2023-10-15
'Solution'

oleksandr-pupchenko:


app/main.py
```diff
@@ -1 +1,38 @@
...
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Carnivore) and not herbivore.hidden:
```
LiudmylaKulinchenko, line 29: You should check here if it is a herbivore

app/main.py
```diff
@@ -1 +1,38 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 28: Fix annotation and parameter name, your function gets not only herbivores here

app/main.py
```diff
@@ -1 +1,38 @@
...
+        if not isinstance(herbivore, Carnivore) and not herbivore.hidden:
+            herbivore.health -= 50
+        if herbivore.health <= 0:
+            Animal.alive.remove(herbivore)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
```
LiudmylaKulinchenko, line 38: define the class before Carnivore to have consistent logic

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #744, opened 2023-10-20
Solution

DenVia:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        self.name = name
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{" \
+               f"Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}" \
+               f"}}"
```
LiudmylaKulinchenko, line 15: Use parentheses instead of `\` as a line breaker

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
LiudmylaKulinchenko, line 24: U can make it easier
Solution: self.hidden = not self.hidden

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if not target.hidden and isinstance(target, Herbivore):
+            target.health -= 50
+
+            if target.health <= 0:
+                for animal in Animal.alive:
+                    if animal.name == target.name:
+                        Animal.alive.remove(target)
```
LiudmylaKulinchenko, line 35: No need to iterate, you can just remove by value

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🚀

PR #745, opened 2023-10-20
Solution

oboliziuk:


app/main.py
```diff
@@ -1 +1,39 @@
...
+            self.hidden = True
+        else:
+            self.hidden = False
+        return self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: "Animal") -> int:
+        if isinstance(other, Herbivore) is True and other.hidden is False:
```
ArsenPidhoretskyi, line 37: No need to write is True/False
```suggestion
        if isinstance(other, Herbivore) and not other.hidden:
```

app/main.py
```diff
@@ -1 +1,39 @@
...
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
+        return self.hidden
```
ArsenPidhoretskyi, line 31: Simplify you logic 
HINT: use not keyword

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
```
ArsenPidhoretskyi, line 28: No need to return self.hidden

ArsenPidhoretskyi: It is redundant due to task requirements

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: "Animal") -> int:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+        return other.health
```
ArsenPidhoretskyi, line 36: No need to return health. It is redundant due to task requirements

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __setattr__(self, attr: str, value: int) -> None:
+        if attr == "health" and value <= 0 and self in Animal.alive:
+            Animal.alive.remove(self)
+        super().__setattr__(attr, value)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
Yurnerroo, line 20: use () instead of \

app/main.py
```diff
@@ -1 +1,36 @@
...
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
```
Yurnerroo, line 26: does it return bool?

app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = not self.hidden
+        return
```
Yurnerroo, line 28: what's this for?

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: "Animal") -> int:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+        return
```
Yurnerroo, line 36: why do you need this return?

Yurnerroo (changes requested):

app/main.py
oboliziuk: Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores.

app/main.py
oboliziuk: None?

app/main.py
```diff
@@ -1 +1,39 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __setattr__(self, attr: str, value: int) -> None:
+        if attr == "health" and value <= 0 and self in Animal.alive:
+            Animal.alive.remove(self)
+        super().__setattr__(attr, value)
```
vsmutok, line 20: why did you override the `__setattr__` method?
it is better to create a separate method for this

vsmutok (changes requested):

PR #746, opened 2023-10-21
Solution

maksimpro001:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+    def __setattr__(self, attr: str, value: int) -> None:
+        if attr == "health" and value <= 0 and self in Animal.alive:
+            Animal.alive.remove(self)
+        super().__setattr__(attr, value)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
YukihiroSM, line 22: Don't use `\` for line separation. 
Better to use `(` `)`

YukihiroSM (changes requested):

PR #749, opened 2023-10-24
Solution

maxvell4:


app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+
+        if not animal.hidden:
```
pavlejviki, line 29:
```suggestion
if isinstance(animal, Herbivore) and not animal.hidden:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+            animal.health -= 50
+            if animal.health <= 0:
+                animal.alive.remove(animal)
+        if isinstance(animal, Carnivore):
```
pavlejviki, line 33: follow the comment above and remove this if block

pavlejviki (changes requested):

app/main.py
maxvell4: Thank`s

Dimosphen1 (approved):
LGTM, well done!

PR #751, opened 2023-10-24
Solution

MykytaPD09:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
pavlejviki, line 7: Don't hardcode the value for self.hidden. add it to parameters

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: int) -> None:
```
Dimosphen1, line 32: Change your type annotation, `herbivore` should not be `int`.

Dimosphen1 (changes requested):
One minor change was requested.

app/main.py
```diff
@@ -1 +1,34 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: "Herbivore") -> None:
```
Dimosphen1, line 32: Do not use quotes `""` in type annotations, declare class directly to indicate that you expect its instance as an input. Use from `__future__ import annotations` in case you face issues with this syntaxis.

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
```
Dimosphen1, line 8: Indicate in type annotation that `health` and `hidden` are optional arguments.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #755, opened 2023-10-27
Implemented classes Animal, Herbivore, Carnivore with appropriate met…

Dikey945:
…hods

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,37 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = False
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 30: fix annotations and parameter name here. If your method takes only Herbivore, then you wouldn't need the check in the next line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 😉☀

PR #758, opened 2023-10-28
Solution

SashaTarnavsky:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
ArsenPidhoretskyi, line 17: Use () instead of \

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+    def __repr__(self) -> str:
+        return self.__str__()
```
ArsenPidhoretskyi, line 20:
```suggestion
        return str(self)
```

ArsenPidhoretskyi (changes requested):

app/main.py
SashaTarnavsky: > Use () instead of \

thanks

ArsenPidhoretskyi (approved):
Well done!

PR #759, opened 2023-10-29
Solution

eLQeR:


app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return "{" + f"Name: {self.name}, Health: {self.health}, " \
+                     f"Hidden: {self.hidden}" + "}"
```
ArsenPidhoretskyi, line 12: 1. Use brackets instead of `\`
2. Why do you use concatenation? Just use f-string

ArsenPidhoretskyi (changes requested):

app/main.py
eLQeR: because if f-string we cannot use single "{" even with "\". Or I dont know only?)

eLQeR: image

eLQeR: if in one row then it is too long for flake8

YukihiroSM: 1. You still can use brackets instead of `\` for line separation
2. Use `{{` and `}}` to make everything in one f-string.

YukihiroSM (changes requested):

app/main.py
eLQeR: You still can use brackets instead of \ for line separation. I don't get what does that mean, which brackets???)

YukihiroSM, line 12:
```suggestion
        return ("{" + f"Name: {self.name}, Health: {self.health}, "
                     f"Hidden: {self.hidden}" + "}")
```

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #763, opened 2023-10-31
'Solution'

yul1valko:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
Dimosphen1, line 18: Reformat your code to place a string inside `()` and avoid line breaking `\`.

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def __repr__(self) -> str:
```
Dimosphen1, line 20: You could define only `__repr__` for this task, just transfer your corrected code from `__str__` here and delete `__str__`.

app/main.py
```diff
@@ -1 +1,41 @@
...
+            self.hidden = False
+        elif self.hidden is False:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if isinstance(victim, Herbivore)\
+                and victim.health > 0 and not victim.hidden:
```
Dimosphen1, line 37: There is no need to check that `victim` has health greater than 0, you could omit it because you redefine health on line 40. Besides that, do not use line breaking `\` in the if condition.

Dimosphen1 (changes requested):

yul1valko:
Have implemented requested changes.

vasylhnatiuk (approved):

PR #770, opened 2023-11-06
Solution

Liliyakhu:


app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+class Animal:
+
```
viktoria-rybenchuk, line 2: there is no need for a blank line between the class definitions and class attributes.

app/main.py
```diff
@@ -1 +1,38 @@
...
+        return (
+            f"{{"
+            f"Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+
```
viktoria-rybenchuk, line 26: fix here according to the comment above

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(inst: Herbivore) -> None:
```
viktoria-rybenchuk, line 34: write descriptive parameter name instead of `inst`

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #772, opened 2023-11-09
'Solution_09112023'

Anton-Konyk:


app/main.py
```diff
@@ -1 +1,31 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if isinstance(animal, Herbivore) and not animal.hidden:
+            animal.health -= 50
+        if animal.health <= 0:
+            Animal.alive.pop(Animal.alive.index(animal))
```
NikOneZ1, line 31: It would be better to use .remove() method

NikOneZ1 (changes requested):

PR #773, opened 2023-11-09
'solution'

boris-devs:


app/main.py
```diff
@@ -1 +1,44 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name:
+                 str, health:
+                 int = 100, hidden:
+                 bool = False) -> None:
```
EdAlekseiev, line 8:
```suggestion
    def __init__(self,
                 name:
                 str, health:
                 int = 100,
                 hidden: bool = False) -> None:
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
EdAlekseiev, line 18: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    @classmethod
+    def del_dead_animal(cls) -> None:
+        cls.alive = [animal for animal in Animal.alive if animal.health > 1]
```
EdAlekseiev, line 22: I think, that animals with 1 health are live too.
```suggestion
        cls.alive = [animal for animal in Animal.alive if animal.health > 0]
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
EdAlekseiev, line 31: You can simplify this code
```suggestion
        self.hidden = not self.hidden
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if animal.hidden is True or isinstance(animal, Carnivore):
+            return
+        else:
+            animal.health -= 50
```
YukihiroSM, line 37:
```suggestion
        if animal.hidden is True or isinstance(animal, Carnivore):
            return
        animal.health -= 50
```

app/main.py
```diff
@@ -1 +1,40 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    @classmethod
+    def del_dead_animal(cls) -> None:
+        cls.alive = [animal for animal in Animal.alive if animal.health > 0]
```
YukihiroSM, line 21: Why do you need this method if you are not using this?

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #774, opened 2023-11-10
Solution for py_herbivores_and_carnivores

VladyslavBazhyn:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
EdAlekseiev, line 18: Use `()` instead of `\`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = True if not self.hidden else False
```
YukihiroSM, line 26:
```suggestion
        self.hidden = not self.hidden
```

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Carnivore(Animal):
+
+    def bite(self, victim: Herbivore) -> None:
+
+        if isinstance(victim, Herbivore) and not victim.hidden:
+            victim.health -= 50
+            if victim.health <= 0:
+                Animal.alive.pop(Animal.alive.index(victim))
```
YukihiroSM, line 36: You can just use list.remove()

YukihiroSM (changes requested):

YukihiroSM (approved):
Nice job!

PR #777, opened 2023-11-14
Done

Moonralex:


app/main.py
```diff
@@ -1 +1,42 @@
...
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
viktoria-rybenchuk, line 32:
```suggestion
        self.hidden = not self.hidden
```

viktoria-rybenchuk: this  approach is more concise and  better because it achieves the same result as the first approach but in a more straightforward manner

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore) -> None:
```
viktoria-rybenchuk, line 38: use descriptive variable name instead of `other`

app/main.py
```diff
@@ -1 +1,42 @@
...
+        if isinstance(other, Herbivore):
+            if not other.hidden:
+                other.health -= 50
```
viktoria-rybenchuk, line 41: you can check `if other.health <= 0` and then remove it from list. In this way you can avoid using  ` is_alive()` function

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Great job!

PR #781, opened 2023-11-20
Solution

KwintDimon:


app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: str = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
pavlejviki, line 7: pass the value for self.hidden as an argument

pavlejviki (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #784, opened 2023-11-21
Solution

Maxboond:


app/main.py
```diff
@@ -1 +1,32 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: int) -> None:
```
Dimosphen1, line 30: Change your type annotation, `target` should not be `int`.

app/main.py
```diff
@@ -1 +1,32 @@
...
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
```
Dimosphen1, line 25: Change your return type annotation, there is no `bool` that is returned.

app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
Dimosphen1, line 7: It is a more conventional way to leave no arguments on the same lines with opening and closing parenthesis in the `__inti__`, place each argument on a new line.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
```
Dimosphen1, line 7: You could indicate in the type annotation that arguments that have default values are optional in this way
```suggestion
            health: int | None = 100,
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #798, opened 2023-11-27
Develop

YevhenKunieiev:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __repr__(self) -> str:
+        prep_msg = "Name: {0}, Health: {1}," \
+                   " Hidden: {2}".format(self.name, self.health, self.hidden)
+        return f"{{{prep_msg}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
```
ArsenPidhoretskyi, line 30: use not keyword instead of if-else

ArsenPidhoretskyi (changes requested):

app/main.py
YevhenKunieiev: like this one?

        if not self.hidden:
            self.hidden = True

but in this case one test is falling down(

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.alive.append(self)
+
+    @classmethod
+    def remove_animal(cls, herbivore_unit: object) -> None:
+        cls.alive.remove(herbivore_unit)
+
+    def __repr__(self) -> str:
+        prep_msg = "Name: {0}, Health: {1}," \
+                   " Hidden: {2}".format(self.name, self.health, self.hidden)
+        return f"{{{prep_msg}}}"
```
vsmutok, line 24:
```suggestion
    def __repr__(self) -> str:
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, "
            f"Hidden: {self.hidden}}}"
        )
```

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.health = health
+        self.hidden = hidden
+        self.adding_to_alive()
+
+    def adding_to_alive(self) -> None:
+        self.alive.append(self)
+
+    @classmethod
+    def remove_animal(cls, herbivore_unit: object) -> None:
+        cls.alive.remove(herbivore_unit)
```
vsmutok, line 19: I think this methods redundant

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health = herbivore.health - 50
+            if herbivore.health <= 0:
+                Animal.remove_animal(herbivore)
```
vsmutok, line 39:
```suggestion
                Animal.remove(herbivore)

File: app/main.py [hunk 5]
```diff
@@ -1 +1,39 @@
...
+
+class Animal:
+    alive: ClassVar = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.adding_to_alive()
```
vsmutok, line 12:
```suggestion
        self.alive.append(self)
```

vsmutok (changes requested):

PR #801, opened 2023-11-30
Solution

OlgaRozd:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore: Herbivore) -> None:
```
Polyakiv-Andrey, line 29: rewrite this method without using return, change condition

Polyakiv-Andrey (changes requested):

PR #808, opened 2023-12-08
Solution

DanuloM:


app/main.py
```diff
@@ -1 +1,30 @@
...
+        self.hidden = False
+        self.health = health
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    def death(self) -> None:
```
pavlejviki, line 15:
```suggestion
    def die(self) -> None:
```

pavlejviki: n general, using a verb as a function name is a good practice in programming because it conveys an action.

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #819, opened 2023-12-19
Solution

aLEKS-e3:


Vladyslav-Pliashechnyk (approved):

app/main.py
```diff
@@ -1 +1,29 @@
...
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name_str = f"Name: {self.name}"
+        health_str = f"Health: {self.health}"
+        hidden_str = f"Hidden: {self.hidden}"
+        return f"{{{name_str}, {health_str}, {hidden_str}}}"
```
NEkropo1, line 16: too complicated, you can just pass them all in a single return of some structure

NEkropo1 (changes requested):
LNGTM!

app/main.py
aLEKS-e3: no easy options!!!

app/main.py
```diff
@@ -1 +1,28 @@
...
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.alive.append(self)
```
NEkropo1, line 10: consider moving this to `__new__`

NEkropo1 (approved):

app/main.py
aLEKS-e3: thank you for the advice, I'll do some research)

PR #843, opened 2024-01-04
Solution

kabust:


ArturPoltser (approved):

app/main.py
```diff
@@ -1 +1,39 @@
...
+            Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        animal = {
+            "Name": self.name,
+            "Health": self.health,
+            "Hidden": self.hidden
+        }
+
+        return str(animal).replace("'", "")
```
NEkropo1, line 22: make it in a single return

app/main.py
```diff
@@ -1 +1,39 @@
...
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if herbivore.hidden or isinstance(herbivore, Carnivore):
+            return
+
+        herbivore.health -= 50
+
+        if herbivore.health <= 0:
+            died_index = super().alive.index(herbivore)
+            del super().alive[died_index]
```
NEkropo1, line 39: super().alive is not a valid way to access the alive attribute of the parent class. Instead, you should access it through the class itself (Animal.alive), as class attributes are shared among all instances of the class.

NEkropo1 (changes requested):

PR #849, opened 2024-01-11
Solution

fposlushnyi:
5.10. OOP Single Inheritance

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,33 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,33 @@
...
+            health: int = 100,
+            hide: bool = False
+    ) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hide
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
```
vasylhnatiuk, line 18: use () instead of \

vasylhnatiuk (changes requested):

PR #853, opened 2024-01-13
Solution

MaxymChyncha:


app/main.py
```diff
@@ -1 +1,33 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
```
kabust, line 30: It's better to check if the condition is false by using negation `if not herbivore.hidden`

kabust (approved):

PR #854, opened 2024-01-13
'Solution'

JohnnyWalker010:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        herbivores = [
+            animal for animal in Animal.alive
+            if isinstance(animal, Herbivore)
+            and not animal.hidden
+        ]
+        if herbivores:
+            target_herbivore = herbivores[0]
+            target_herbivore.health -= 50
+            if target_herbivore.health <= 0:
+                target_herbivore.die()
```
barabarinov, line 41: Task description says that the bite method should receive Herbivore instance as an argument :)

barabarinov (changes requested):

PR #858, opened 2024-01-14
Solution

yulias97:


app/main.py
```diff
@@ -1 +1,32 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
+        herbivore.health -= 50
```
PythonZem, line 30: I think that there is no need to return here. It is necessary to check that it is a herbivore and that it does not hide. When the check is done, reduce the health

PythonZem (approved):
nice code!

app/main.py
yulias97: thanks

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore) or herbivore.hidden:
+            return
```
NEkropo1, line 29:
```suggestion
        if not isinstance(herbivore, Herbivore) or herbivore.hidden:
            return

```

NEkropo1 (approved):
Good job!

PR #861, opened 2024-01-14
Solution

oksankakot:


mdubyna (approved):
Great job!

app/main.py
```diff
@@ -1 +1,33 @@
...
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False)\
+            -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
NEkropo1, line 11: best approach would move this to `__new__`, but not obligatory.

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.die()
```
NEkropo1, line 33: You won't need that in this exact code, but overall logic tells us that we will make this check often, so best approach would move this to some static method.

NEkropo1 (approved):
Good job!

PR #862, opened 2024-01-15
Solution herbivores and carnivores

mdubyna:


barabarinov (approved):

app/main.py
```diff
@@ -1 +1,32 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return ("{Name: " + self.name + ", Health: " + str(self.health)
+                + ", Hidden: " + str(self.hidden) + "}")
```
NEkropo1, line 15: use f-strings )

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if not isinstance(other, Carnivore) and not other.hidden:
+            if other.health - 50 < 1:
+                Animal.alive.remove(other)
+            else:
+                other.health -= 50
```
NEkropo1, line 32: you could a little simplify this by removing health at first, and make check at second.

NEkropo1 (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if not isinstance(other, Carnivore) and not other.hidden:
+            other.health -= 50
+            if other.health - 50 < 0:
```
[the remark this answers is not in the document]
mdubyna: if I follow this example I get an error
LQ_error

PR #865, opened 2024-01-16
Solution

VladKyianytsia:


Bulleter06 (approved):
It is very comfortable code for watching, awesome!

PR #871, opened 2024-01-16
'Solution'

Nick-r2c:


AntonSkyrda (approved):
Looks good

app/main.py
```diff
@@ -12,7 +13,7 @@ def __init__(
         self.hidden = hidden
         Animal.alive.append(self)
 
-    def is_alive(self) -> None:
+    def keep_alive(self) -> None:
```
NEkropo1, line 16: keep alive makes animal dead(removes from list)? xD
sry, but you need to try more )

NEkropo1 (changes requested):

app/main.py
```diff
@@ -13,7 +13,7 @@ def __init__(
         self.hidden = hidden
         Animal.alive.append(self)
 
-    def keep_alive(self) -> None:
+    def remove_expired_animals(self) -> None:
```
NEkropo1, line 16: `remove_dead_animal`
I don't think there's some expiration date here)
or, maybe something like this, for total explicity:
`remove_from_alive_if_dead`

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

PR #875, opened 2024-01-16
Solution

artemgrishko:


artemlitvenko (approved):

app/main.py
```diff
@@ -17,13 +17,12 @@ def die(self) -> None:
...
-                "Name: " + str(self.name) + ", " +
-                "Health: " + str(self.health) + ", " +
-                "Hidden: " + str(self.hidden) +
-                "}"
-        )
+        return ("{"
+                + "Name: " + str(self.name) + ", "
+                + "Health: " + str(self.health) + ", "
+                + "Hidden: " + str(self.hidden)
+                + "}")
```
NEkropo1, line 24: I guess comment was that you don't need f-strinf in every lane, so you can use f-strings where needed, but pass regular strings in other cases.

NEkropo1 (changes requested):

NEkropo1 (approved):
Good job!

PR #877, opened 2024-01-16
'Solution'

artemlitvenko:


Nick-r2c (approved):
GJ!

app/main.py
```diff
@@ -12,27 +9,26 @@ def __init__(
...
         Animal.alive.remove(self)
 
     def __repr__(self) -> str:
-        return f"{{Name: {self.name}, Health: \
-{self.health}, Hidden: {self.hidden}}}"
+        return (f"{{"
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}"
+                f"}}")
```
NEkropo1, line 20: you don't need to use f-strings on every lane, remove them if not necessary.

NEkropo1 (changes requested):

PR #878, opened 2024-01-16
Solution

tskopenko:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
```
isikach, line 4: It will looks better if you start all parameters from new line
https://github.com/mate-academy/py-distance-class/pull/930#discussion_r1449982199

isikach (approved):

PR #882, opened 2024-01-17
Solution1

Dmitriy1504:


app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from __future__ import annotations
+from typing import Union
```
Dimosphen1, line 2: Do not use `Union`, use `|` in the type annotation instead in the future.

Dimosphen1 (changes requested):
LGTM, one comment is left for consideration.

Dimosphen1 (approved):

PR #885, opened 2024-01-20
Solution

nicksetrakov:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self.health > 0:
+            self.alive.append(self)
+
+    def check_alive(self) -> None:
+        if self.health <= 0:
+            self.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
LiudmylaKulinchenko, line 22: Don't use ` \` as a linebreaker, use parentheses instead

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 36: fix annotation and parameter name here. If your method takes only Herbivore, then you wouldn't need the check in the next line

app/main.py
```diff
@@ -1 +1,39 @@
...
+        return f"{{Name: {self.name}, Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
LiudmylaKulinchenko, line 30: U can use `self.hidden = not self.hidden` instead

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
LGTM!

PR #889, opened 2024-01-24
Solution

AlexGrytsai:


app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
EdAlekseiev, line 5: If you separate arguments to a few lines, each argument must be in a separate line
```suggestion
    def __init__(
        self,
        name: str,
        health: int = 100,
        hidden: bool = False
    ) -> None:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #900, opened 2024-01-29
solution

prev1ew:


prev1ew:
image
this works, but there's a problem with tests

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.health = health
+        self.hidden = hidden
+        if health > 0:
+            Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        object_items_str = ", ".join(
+            (f"{k.capitalize()}: {v}" for k, v in vars(self).items())
+        )
+        return f"{{{object_items_str}}}"
```
Dimosphen1, line 20: Write `__repr__` method using one f-string and defining `Name`, `Health`, and `Hidden` directly without using additional cycles. It would help to improve the readability.

app/main.py
```diff
@@ -1 +1,39 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, victim: Herbivore) -> None:
+        if (
+                victim.health > 0
+                and isinstance(victim, Herbivore)
```
Dimosphen1, line 34: It would be a better approach to check first that the `victim` is an instance of the `Herbivore` class and after that conduct additional checks that rely on its attributes to prevent potential errors in case other objects would be passed to this method.

Dimosphen1 (changes requested):

app/main.py
prev1ew: It wont pass tests, its ok?

Dimosphen1 (approved):
LGTM, well done!

PR #901, opened 2024-01-29
solution

DenysST:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
```
LiudmylaKulinchenko, line 28: type annotation here should not be Herbivore. If it will take only herbivores, we won't need the check in the next line

LiudmylaKulinchenko (approved):

PR #906, opened 2024-02-01
'solution'

liudmylaperchevska:


app/main.py
```diff
@@ -1 +1,36 @@
...
+            "Name": self.name,
+            "Health": self.health,
+            "Hidden": self.hidden
+        }
+        rep = (f"{{"
+               f'Name: {class_dict["Name"]}, '
+               f'Health: {class_dict["Health"]}, '
+               f'Hidden: {class_dict["Hidden"]}'
+               f"}}")
+        return rep
```
pavlejviki, line 22: try to refactor this method in  a more more concise way. avoid creating additional dictionary and variable. simply return the required string

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, whom: Herbivore) -> None:
+        if (not whom.hidden) and isinstance(whom, Herbivore):
+            whom.health -= 50
```
pavlejviki, line 33: whom is not a clear variable name

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
pavlejviki, line 7: the hidden attribute is likely to be set to different values at the creation time of the object, it is better to allow it to be set via the constructor parameters and there you can add a default value False.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #910, opened 2024-02-03
Solution

vsupruniuk:


app/main.py
```diff
@@ -1 +1,32 @@
...
+
+        self.health = health
+        self.hidden = False
+
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                + f"Health: {self.health}, "
+                + f"Hidden: {self.hidden}}}")
```
LiudmylaKulinchenko, line 18: no need to use concatenation, you can remove `+` and it will work the same

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
💪💪 well done

PR #913, opened 2024-02-05
added solution

MAFIOZNIK1999:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        name = "{Name: " + str(self.name)
+        health = ", Health: " + str(self.health)
+        hidden = ", Hidden: " + str(self.hidden) + "}"
+        return name + health + hidden
```
LiudmylaKulinchenko, line 24: explicitly defining the string that should be returned will improve the readability
Just use f-strings and parentheses to split into lines

app/main.py
```diff
@@ -1 +1,41 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Union[Herbivore, Carnivore]) -> None:
```
LiudmylaKulinchenko, line 36: You can use Animal class for annotations here, it will include Herbivore, Carnivore and other classes, that could be created in the future

app/main.py
```diff
@@ -1 +1,41 @@
...
+        if animal.hidden is False and isinstance(animal, Herbivore):
```
LiudmylaKulinchenko, line 37: You can use `not` in the condition
`if not animal.hidden`

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great 💪

PR #926, opened 2024-02-10
add task solution

EugeneSherepa:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{" \
+               f"Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}" \
+               f"}}"
```
LiudmylaKulinchenko, line 20: Using parentheses for string continuation in Python is generally preferred over using the backslash

LiudmylaKulinchenko (changes requested):

PR #927, opened 2024-02-11
Solution added

CreepyMo:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, other: Animal) -> None:
```
Dimosphen1, line 27: Indicate that you expect to obtain exactly `Herbivore` class instance for `other`, not `Animal` class instance in general.

app/main.py
```diff
@@ -1 +1,31 @@
...
+        if not isinstance(other, Carnivore) and not other.hidden:
```
Dimosphen1, line 28: Check that other is exactly `Herbivore` according to the task requirement.

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False) \
```
Dimosphen1, line 4: Place each argument in the `__init__` on the new line including `self` so there will be no arguments on the same line with opening and closing parenthesis. It will help to improve the readability and avoid using line breaking.

Dimosphen1 (changes requested):

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

Polyakiv-Andrey (approved):
Good job!

PR #931, opened 2024-02-12
'Solution'

donbas777:


app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
Dimosphen1, line 6: Place each argument on a new line including `self`, so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability.

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+        return None
```
Dimosphen1, line 21: There is no need to specifically declare that you return `None`, you should omit this line since it is a default behavior.

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore):
+            if animal.hidden is False:
+                animal.health -= 50
+            if animal.health <= 0:
+                Animal.alive.remove(animal)
+        return None
```
Dimosphen1, line 32: There is no need to specifically declare that you return `None`, you should omit this line since it is a default behavior.

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+        return None
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        if isinstance(animal, Herbivore):
+            if animal.hidden is False:
```
Dimosphen1, line 28: Join these two if statements into one using `and` to avoid nesting.

Dimosphen1, line 28: There is no need to explicitly check that `animal.hidden is False`, you could evaluate its boolean type instead
```suggestion
        if isinstance(animal, Herbivore):
            if not animal.hidden:
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
```
Dimosphen1, line 4: Place `self` on a new line so that there would be no arguments on the same line with an opening parenthesis as there are no arguments on the same line with a closing parenthesis.

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #939, opened 2024-02-15
Solution

olga-kim7:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name:"
+                f" {self.name},"
+                f" Health: {self.health},"
+                f" Hidden: {self.hidden}}}"
+                )
```
Dimosphen1, line 20: In case you do not place any code on the same line with a closing parenthesis, you should not place the part of the string on the same line with the opening parenthesis. Besides that, correct your indentation so it would be seen that the string is inside of parentheses and not on the same indentation level.

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
```
Dimosphen1, line 30: You should accept only `Herbivore` for biting according to the task description, that is why you should directly check that. There could be another class added later that could impact the logic in its current form, so change it to 
```suggestion
        if not isinstance(herbivore, Herbivore) or herbivore.hidden:
```

Dimosphen1 (changes requested):

app/main.py
olga-kim7: Все виправлено) дякую

PR #942, opened 2024-02-19
Solution

twonumberfortyfives:


app/main.py
```diff
@@ -1 +1,36 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Animal) -> None:
```
Dimosphen1, line 30: Indicate explicitly in the type annotation that you expect to obtain an instance of `Herbivore` class for `target`

app/main.py
```diff
@@ -1 +1,36 @@
...
+        if isinstance(target, Herbivore) and target.hidden is False:
```
Dimosphen1, line 31: You could simplify this check
```suggestion
        if isinstance(target, Herbivore) and not target.hidden:
```

app/main.py
```diff
@@ -1 +1,36 @@
...
+            target.health -= 50
+            if target.health <= 0:
+                target.die()
+        else:
+            return
```
Dimosphen1, line 36: There is no need for this code, remove it

Dimosphen1 (changes requested):

PR #945, opened 2024-02-21
implement 3 classes Carnivore/Herbivore/Animal

MaxSEngineer:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def bite(cls, obj: Herbivore) -> None:
+        for animal in super().alive:
+            if (
+                animal.name == obj.name
+                and isinstance(animal, Herbivore)
+                and not animal.hidden
+            ):
+                animal.health -= 50
+            if animal.health <= 0:
+                super().alive.remove(animal)
```
vsmutok, line 43:
```suggestion
    @staticmethod
    def bite(other_animal: Herbivore) -> None:
        if isinstance(other_animal, Herbivore) and not other_animal.hidden:
            other_animal.health -= 50
            other_animal.check_if_alive()
```

app/main.py
```diff
@@ -1 +1,43 @@
...
+        )
+
+    def __str__(self) -> str:
+        return Animal.alive
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = True if not self.hidden else False
```
vsmutok, line 28:
```suggestion
        self.hidden = not self.hidden
```

vsmutok (changes requested):

app/main.py
MaxSEngineer: Thanks, its really more clear code, i don't know why i used loop)

app/main.py
MaxSEngineer: Makes sense so in short thanks

PR #963, opened 2024-02-27
Solution

Sergunshot:


app/main.py
```diff
@@ -1 +1,29 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore_animal: Herbivore) -> None:
+        if isinstance(herbivore_animal, Herbivore) and \
+                not herbivore_animal.hidden:
```
EdAlekseiev, line 26: Use `()` instead of `\`

EdAlekseiev (changes requested):

app/main.py
Sergunshot: Done

EdAlekseiev (approved):
Good job

PR #976, opened 2024-03-06
done_1

Mirinness:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name: str = name
+        self.health: int = health
+        self.hidden: bool = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, " \
+               f"Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
vsmutok, line 16: better to use `()` then `\`

app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+from typing import List
+
+
+class Animal:
+    alive: List["Animal"] = []
```
vsmutok, line 5:
```suggestion
    alive: List[Animal] = []
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name: str = name
+        self.health: int = health
+        self.hidden: bool = False
+        Animal.alive.append(self)
```
vsmutok, line 11:
```suggestion
    def __init__(self, name: str, health: int = 100, hidden: bool = False) -> None:
        self.name: str = name
        self.health: int = health
        self.hidden: bool = hidden
        Animal.alive.append(self)
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: "Herbivore") -> None:
+        if isinstance(herbivore, Carnivore) or herbivore.hidden:
+            return
+        herbivore.health -= 50
+        if herbivore.health <= 0:
+            herbivore.die()
```
vsmutok, line 34:
```suggestion
    @staticmethod
    def bite(herbivore: Herbivore) -> None:
        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
            herbivore.health -= 50
            if herbivore.health <= 0:
                Animal.alive.remove(herbivore)
```

vsmutok (changes requested):

app/main.py
Mirinness: For some reason, he does not want to accept "Animal" without legs

Mirinness:
done

PR #980, opened 2024-03-10
Solution

Bujhm99:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, obj: Herbivore) -> None:
```
LiudmylaKulinchenko, line 25: Type annotaton is not correct, since you could get eather Herbivore or Carnivore here. Otherwise why would you need a check in the next line?
Use parent class for annotations here

LiudmylaKulinchenko, line 25: Use descriptive name for a parameter. Anythin in a python is an object

app/main.py
```diff
@@ -1 +1,31 @@
...
+        if isinstance(obj, Herbivore) and not obj.hidden:
+            obj.health -= 50
+            # remove from alive and fix health
```
LiudmylaKulinchenko, line 28: Consider to remove this comment. The logic is clear without it, what do you think?

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ!

PR #984, opened 2024-03-13
Solution

Kate-Kotushenko:


app/main.py
```diff
@@ -1 +1,36 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> bool:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
vsmutok, line 26: better to use `not` is this case

app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+from __future__ import annotations
```
vsmutok, line 1: remove redundant import

vsmutok (changes requested):

app/main.py
Kate-Kotushenko: Done

app/main.py
Kate-Kotushenko: 31  def bite(self, animal: Herbivore | Carnivore) -> int | list:
For this annotation I've used from __future__ import annotations

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def hide(self) -> bool:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Herbivore | Carnivore) -> int | list:
```
vasylhnatiuk, line 31: this  returns None

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> bool:
```
vasylhnatiuk, line 22:
```suggestion
    def hide(self) -> None:
```

vasylhnatiuk (changes requested):

app/main.py
Kate-Kotushenko: Sorry, I was inattentive

app/main.py
Kate-Kotushenko: Done

vasylhnatiuk (approved):

PR #988, opened 2024-03-17
Solution for task herbivores and carnivores

oleksandra-shershen:


app/main.py
```diff
@@ -1 +1,34 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
nicksetrakov, line 30: it's staticmethod because you don't use self in this method and  here I'd had the same problem with annotation
image

nicksetrakov (changes requested):

app/main.py
oleksandra-shershen: Done

nicksetrakov (approved):

app/main.py
```diff
@@ -1 +1,35 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def die(self) -> None:
+        self.alive.remove(self)
```
YuriiKindrat, line 16: use class name for entering class attr

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
YuriiKindrat, line 13: same is here

YuriiKindrat (changes requested):

Oleksl888 (approved):
Good!

PR #1000, opened 2024-03-20
Solution

panicua:


SashaKisliy (approved):
GJ!

app/main.py
```diff
@@ -1 +1,30 @@
...
+class Animal:
+    alive = []
+
+    def __init__(self, name: str,
+                 health: int = 100, hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        self.__class__.alive.append(self)
```
YuriiKindrat, line 10: Use Class name instead of ` self.__class__.` here

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if not animal.hidden and animal.__class__ != self.__class__:
```
YuriiKindrat, line 27: same is here )

YuriiKindrat (changes requested):

app/main.py
panicua: why is it bad to use ```self.__class__```? What if we have a huge class we have scroll back at it's name?

YuriiKindrat: it is common practice to use class names for working with some class attrs/methods for better readability

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if not animal.hidden and animal.__class__ != Carnivore:
```
YuriiKindrat, line 28: could u use isinstance method here ? instead of 
`animal.__class__!= Carnivore`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    def __init__(self, name: str,
+                 health: int = 100, hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
```
Oleksl888, line 12: using mutiline string would make your code more readable in this case.

```
    def __repr__(self) -> str:
        return f"""
        {{
        Name: {self.name}, 
        Health: {self.health},
        Hidden: {self.hidden}
        }}
        """

Oleksl888 (approved):
well done sir!

PR #1001, opened 2024-03-21
Solution Herb_Carn March 21 ver 1.0

DmytroPieskov:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal_instance: Animal) -> None:
```
Dimosphen1, line 24: Indicate that you expect to obtain `Herbivore` class instance here explicitly in the type annotation

Dimosphen1, line 24: It would be better to use `animal` variable name instead of `animal_instance` here

app/main.py
```diff
@@ -1 +1,31 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> bool:
```
Dimosphen1, line 17: There is no need to return `bool` value here, return `None` in the type annotation and just perform hide operation in the body without explicit return statement

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal_instance: Animal) -> None:
+        if (
+            isinstance(animal_instance, Herbivore)
+            and not animal_instance.hidden
+        ):
+            animal_instance.health = animal_instance.health - 50
```
Dimosphen1, line 29: Use `-=` here to perform this operation

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1002, opened 2024-03-21
Develop

Nyxan:


app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore):
+            return print("Carnivores cannot bite carnivor")
+        if herbivore.hidden:
+            return print("Cannot bite a hidden herbivore")
```
Oleksl888, line 30: printing is not part of the task and besides print function returns `None` so you can simply make a single check like so:
`if not herbivore.hidden and not isinstance(herbivore, Carnivore):`

Oleksl888 (approved):
well done!

PR #1008, opened 2024-03-21
Solution

SerhiiMar:


app/main.py
```diff
@@ -1 +1,30 @@
...
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.__dict__['name']}, "
+                + f"Health: {self.__dict__['health']}, "
+                + f"Hidden: {self.__dict__['hidden']}}}")
```
yossosh, line 17: Use interpolation instead of concatenation will be better. You can find this advice in checklist.md

yossosh (changes requested):
Use interpolation instead of concatenation

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.__dict__['name']}, "
+                f"Health: {self.__dict__['health']}, "
+                f"Hidden: {self.__dict__['hidden']}}}")
```
YuriiKindrat, line 17: why don't u use just self.name etc. here?)

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.__class__.alive.append(self)
```
YuriiKindrat, line 12: use class name to access class attrs

YuriiKindrat (changes requested):

app/main.py
SerhiiMar: yeah)

app/main.py
```diff
@@ -1 +1,31 @@
...
+                 health: int = 100,
+                 hidden: bool = False) -> None:
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
```
Oleksl888, line 15: You can also use triple quotes for multi line comments to make it more elegant :)

```
def __repr__(self) -> str:
        return f"""
        {{Name: {self.name}, 
        Health: {self.health},
        Hidden: {self.hidden}}}
        """

Oleksl888 (approved):
Good Job!

PR #1009, opened 2024-03-21
Solution for herbivores and carnivores

mikhailyemets:


SerhiiMar (approved):
I don't see what can be improved. Good job!

app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        self.alive.append(self)
```
YuriiKindrat, line 13: u should use class name for accessing class attrs

app/main.py
```diff
@@ -1 +1,33 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if not animal.hidden and animal.__class__ != self.__class__:
+            animal.health -= 50
+        if animal.health < 1:
+            self.alive.remove(animal)
```
YuriiKindrat, line 33: same is here

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
+        if not animal.hidden and animal.__class__ != self.__class__:
```
YuriiKindrat, line 30: use isinstance method here insted of this `animal.__class__ != self.__class__`

YuriiKindrat: also it is better to use class name instead of self.__class__

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
```
Oleksl888, line 16: You could use mutiline string for readablilty 
```
def __repr__(self) -> str:
        return f"""
        {{
        Name: {self.name}, 
        Health: {self.health},
        Hidden: {self.hidden}
        }}
        """

Oleksl888 (approved):
good job!

PR #1013, opened 2024-03-22
Solution

siaserb:


alonasorochynska (changes requested):
It seems that the "bite" method never uses the "self" argument. What about using some decorator to declare this?

YuriiKindrat (approved):
LGTM!

PR #1017, opened 2024-03-23
Solution

Ruxxell54:


YuriiKindrat (approved):
GJ
but it is better to place one arg/same number of args per line in init method

PR #1019, opened 2024-03-24
Implement Animal, Herbivore and Carnivore classes

bwaah1:


dkotkod (approved):
all tests passed. good job

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore_animal: Herbivore) -> None:
+        if not isinstance(herbivore_animal, Carnivore):
+            if not herbivore_animal.hidden:
```
Oleksl888, line 28: you can write this in one line using `and`

app/main.py
```diff
@@ -1 +1,34 @@
...
+                herbivore_animal.health -= 50
+        if herbivore_animal.health <= 0:
```
Oleksl888, line 30: this condition should be checked only if previous went through and animal health was reduced

app/main.py
```diff
@@ -1 +1,34 @@
...
+            Animal.alive = [animal
```
Oleksl888, line 31: you don't need to create new list to remove a dead animal. think which list method you can use for that to keep it simple

Oleksl888 (changes requested):

Oleksl888 (changes requested):
please change

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore_animal: Herbivore) -> None:
+        if (not isinstance(herbivore_animal, Carnivore)
+                and not herbivore_animal.hidden):
+            herbivore_animal.health -= 50
+            if herbivore_animal.health <= 0:
+                for animal in Animal.alive:
+                    if animal.name == herbivore_animal.name:
+                        Animal.alive.remove(animal)
```
Oleksl888, line 33: This is overcomplicated. `Animal.alive` contains is a  list of Animal instances. On line 30 you check if `herbivore_animal` is dead, so what would be the next step?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore_animal: Herbivore) -> None:
+        if isinstance(herbivore_animal, Carnivore) or herbivore_animal.hidden:
+            return
+
+        herbivore_animal.health -= 50
+        if herbivore_animal.health <= 0:
+            Animal.alive = [animal for animal in Animal.alive
```
Oleksl888, line 32: can you simply use `remove` method here?

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore_animal: Herbivore) -> None:
+        if isinstance(herbivore_animal, Carnivore) or herbivore_animal.hidden:
```
Oleksl888, line 27: you can avoid writing a return here by reversing the logic:
`if not isinstance(...) and not ...: -> then do the below logic`

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #1021, opened 2024-03-25
'Solution'

dkotkod:


app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore):
+            return print("Carnivores cannot bite carnivor")
+        if herbivore.hidden:
+            return print("Cannot bite a hidden herbivore")
```
YuriiKindrat, line 30: no need to return print, u could just print this )

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore):
+            print("Carnivores cannot bite carnivor")
```
Oleksl888, line 28: There are no requirements for printing anything - please remove here and below. 
Also Think how you can combine two if statements on line 27 and 30 to simplify your code and improve readability

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore):
+            return
+        elif herbivore.hidden:
+            return
```
Oleksl888, line 30: Don't get into a habit of making extra `if-elif` and `return` statements. Here you can simply try opposite logic:
`if not first condition and not second condition  -> do something`

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Carnivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                print(f"{herbivore.name} is dead")
```
Oleksl888, line 30: I will approve, but you need to be sure that you fix all comments that a reviewer left you before you push your changes. In real project things like this will have negative impact on your evaluation at work

PR #1026, opened 2024-03-26
'Solution'

March1205:


vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1 +1,41 @@
...
+                f" Health: {self.health}, Hidden: {self.hidden}}}")
+
+    def die(self) -> None:
+        Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+        if self.health <= 0:
```
vasylhnatiuk, line 26: you don't need this check here, only in bite()

app/main.py
```diff
@@ -1 +1,41 @@
...
+from typing import Optional
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: Optional[int] = None) -> str:
+        self.name = name
+        self.health = health
+        if self.health is None:
```
vasylhnatiuk, line 10: you can avoid it, write 100 as default atgument in init

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1038, opened 2024-04-01
solution

ArtemLeo:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if isinstance(target, Carnivore) or target.hidden:
```
Dimosphen1, line 40: Rever your condition to check that the passed animal is actually Herbivore and it is not hidden so that your logic woultd not break with adding new animal subclasses

app/main.py
```diff
@@ -1 +1,46 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    def take_damage(self, damage: int) -> None:
+        self.health -= damage
+        if self.health <= 0:
+            self.health = 0
+            self.die()
```
Dimosphen1, line 26: This method is not used, you could remove it

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def die(self) -> None:
+        Animal.alive.remove(self)
+        print(f"{self.name} has died.")
```
Dimosphen1, line 30: There is no need to add redundant print statements, remove them from the code

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
```
Dimosphen1, line 39: Be more specific in a type annotation, indicate that target should be a `Herbivore` class instance

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
...
+        if not isinstance(target, Herbivore) or target.hidden:
+            print("Cannot bite!")
```
pavlopro, line 34: Remove redundant print statement

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1050, opened 2024-04-10
implemet Herbivores and carnivores

daikj88:


app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
Dimosphen1, line 7: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

app/main.py
```diff
@@ -1 +1,33 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if not (isinstance(other, Carnivore) or other.hidden):
```
Dimosphen1, line 30: You check should be explicit and not based on assumption that if an instance is not `Carnivore`, then it is `Herbivore`, you should explicitly check that `other` is `Herbivore`. Other classes could be added later and `bite` method should preserve working only with `Herbivore` class instances

Dimosphen1 (changes requested):

vsmutok (approved):
Looks Good!

PR #1059, opened 2024-04-19
Solution

VlaBeh:


VlaBeh:
Why do I can't try pytest?

vsmutok (changes requested):
tests Failed!

VlaBeh:
class Animal:
    alive = []

    def __init__(self, name: str, health: int = 100,
        self.name = name
        self.health = health
        self.hidden = hidden
        Animal.alive.append(self)

    def __repr__(self) -> str:
        return (f"{{Name: {self.name}, Health: {self.health}, "
                f"Hidden: {self.hidden}}}")

    def die(self) -> None:
        Animal.alive.remove(self)


class Herbivore(Animal):
    def hide(self) -> None:
        self.hidden = not self.hidden


class Carnivore(Animal):
    def bite(self, target: int) -> None:
            target.health -= 50
            if target.health <= 0:
                target.die()
This is my code and it have passed all the tests but I can't do the pull request again.

vsmutok (changes requested):
tests Failed!

PR #1068, opened 2024-04-26
triukhan solution

triukhan:


app/main.py
```diff
@@ -1 +1,31 @@
...
+                 ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
```
LiudmylaKulinchenko, line 17: Don't use `\` as a linebreaker, u can use brackets instead of linebreaker here

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
```
LiudmylaKulinchenko, line 27: fix annotation here. If your method takes only Herbivore, then you wouldn't need the check in the next line

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
```
LiudmylaKulinchenko, line 8: fix formatting

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 💪

PR #1069, opened 2024-05-03
Solution

Shayba228:


app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+from typing import List
+
+
+class Animal:
+    alive: List["Animal"] = []
```
```suggestion
    alive: List[Animal] = []
```

vasylhnatiuk: use from __future__ import annotations

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1082, opened 2024-05-20
Solution

SuskyiVolodymyr:
I don't know is it ok to do line break in __repr__ like this

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -7,7 +7,6 @@ class Animal:
             health: int = 100,
             hidden: bool = False
     ) -> None:
-
         self.name = name
         self.health = health
         self.hidden = hidden
@@ -15,11 +14,13 @@ class Animal:
             Animal.alive.append(self)
 
     def __repr__(self) -> str:
-        return (f"{{"
-                f"Name: {self.name}, "
-                f"Health: {self.health}, "
-                f"Hidden: {self.hidden}"
-                f"}}")
+        return (
+            f"{{"
+            f"Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}"
+            f"}}"
+        )
 
     @staticmethod
     def remove_dead_animals() -> None:
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,41 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,41 @@
...
+    ) -> None:
+
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        if self.health > 0:
+            Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{"
```
DevSpaciX, line 18: Maybe we can try this ?

```
    return (
        f"Name: {self.name}, "
        f"Health: {self.health}, "
        f"Hidden: {self.hidden}"
    )
```
Or if '{{ }}' is necessary ->
```
    return (
        f"{{"
        f"Name: {self.name}, "
        f"Health: {self.health}, "
        f"Hidden: {self.hidden}"
        f"}}"
    )
```

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+class Animal:
+    alive = []
```
Oleg-s-s, line 2: You can use `set` for to leverage faster membership checks and removals.

app/main.py
```diff
@@ -1 +1,42 @@
...
+            f"{{"
+            f"Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}"
+            f"}}"
+        )
+
+    @staticmethod
+    def remove_dead_animals() -> None:
+        Animal.alive = [animal for animal in Animal.alive if animal.health > 0]
```
Oleg-s-s, line 27: Going through all animals is not efficient, you can make something like `change_health` method that will check and change Animal health, and if Animal is dead remove it

Oleg-s-s (changes requested):

Oleg-s-s (changes requested):
.

app/main.py
SuskyiVolodymyr: I can`t, because in test file i have this test:

image

SuskyiVolodymyr: and more other checks where we need to use list

Oleg-s-s: ok

Oleg-s-s (approved):
👍

PR #1096, opened 2024-05-25
'Solution'

BohdanKuzik:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def die(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
```
Oleg-s-s, line 19: Bad approach, You should better remove single animal who is dead, not recreating whole list

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.health = 0
+                Animal.die()
```
Oleg-s-s, line 33: No need to use instance method here, you are not using self

app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
+    def bite(cls, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.health = 0
```
DevSpaciX, line 35: Look , you just can remove the animal from alive animal list instead of recreating new list

app/main.py
```diff
@@ -1 +1,36 @@
...
+                cls.alive.remove(animal)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
```
DevSpaciX, line 30: You can do this method static

app/main.py
```diff
@@ -1 +1,36 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def die(cls) -> None:
+        for animal in cls.alive[:]:
+            if animal.health <= 0:
+                cls.alive.remove(animal)
```
DevSpaciX, line 21: You can remove this method

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,30 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.health = 0
```
DevSpaciX, line 29: That cmove is not necessary )

DevSpaciX (approved):
Good !

PR #1098, opened 2024-05-26
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+    def __init__(self, name: str) -> None:
+        self.name: str = name
+        self.health: int = 100
+        self._hidden: bool = False
+        Animal.alive.append(self)
+
+    @property
+    def hidden(self) -> bool:
+        return self._hidden
```
Oleg-s-s, line 15: What is the purpose of creating `_hidden` and then accessing through `hidden`?)
I dont see in the task that we need private hidden attr

Oleg-s-s (changes requested):

tests/test_main.py
```diff
@@ -144,13 +144,17 @@ def test_print_animal_alive():
...
+    animal = Herbivore("Rex")
+    assert animal.name == "Rex"
+    assert animal.health == 100
+    assert animal.hidden is False
+    assert len(Animal.alive) == 1
+
+    for _ in range(3):
+        Carnivore.bite(animal)
+
+    assert len(Animal.alive) == 0
```
DevSpaciX, line 157: `Never` change the test code, please return everything as it was

app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: "Herbivore") -> None:
```
DevSpaciX, line 31: Why do you using " " here ?)

app/main.py
```diff
@@ -1 +1,35 @@
...
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.die()
```
DevSpaciX, line 35: You just can call the remove method here and it will work I guess

DevSpaciX (changes requested):

PR #1100, opened 2024-05-26
Solution

SrMusienko:


app/main.py
```diff
@@ -1 +1,35 @@
...
+            Animal.alive.remove(self)
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if victim.hidden or not isinstance(victim, Herbivore):
+            return
+        victim.health -= 50
+        if victim.health <= 0:
```
DevSpaciX, line 34: Is that check is duplicated ? ( I am about line 24 ) If so just call remove method instead of creating new func : )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1105, opened 2024-05-27
Solution

lilarin:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Herbivore) -> None:
+        if not victim.hidden and isinstance(victim, Herbivore):
+            victim.health -= 50
+            if victim.health <= 0:
+                Animal.alive.remove(victim)
```
DevSpaciX, line 35: Large nesting of code is bad practice, you can just do a `return` and limit it to nesting at 1 level

DevSpaciX (approved):
Approve , but read the comment please : )

PR #1108, opened 2024-05-27
Sol 1

Unlie9:


app/main.py
```diff
@@ -1 +1,41 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
```
DevSpaciX, line 32: You can add `or` param and check if the `herbivore` is hidden cuz we do not need to change health in that case , yes ?

app/main.py
```diff
@@ -1 +1,41 @@
...
+            return
+
+        if herbivore.hidden is False:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.die()
+
+        if herbivore.hidden is True:
```
DevSpaciX, line 40: So we just passing the default param ?

app/main.py
```diff
@@ -1 +1,41 @@
...
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
+            return
+
+        if herbivore.hidden is False:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                herbivore.die()
```
DevSpaciX, line 38: Not bad idea , but could we just call remove method here ?

DevSpaciX (changes requested):

app/main.py
DevSpaciX: @Unlie9 I think you forget about this

app/main.py
DevSpaciX: I was about to catch this case and do nothing cuz this is default params : )

app/main.py
```diff
@@ -1 +1,39 @@
...
+            return
+
+        if herbivore.hidden:
+            herbivore.health = 100
+        else:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                Animal.alive.remove(herbivore)
+
+        if herbivore.hidden:
```
DevSpaciX, line 38: I was about to catch this case and do nothing cuz this is default params : )

DevSpaciX (changes requested):

app/main.py
Unlie9: i got you

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
```
DevSpaciX, line 28: ` if not isinstance(herbivore, Herbivore) or herbivore.hidden:`
    `return`

app/main.py
```diff
@@ -1 +1,36 @@
...
+            return
+
+        if herbivore.hidden:
+            herbivore.health = 100
```
DevSpaciX, line 32: so now you can delete this code

DevSpaciX (approved):

PR #1109, opened 2024-05-27
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1 +1,31 @@
...
+            Animal.alive.remove(self)
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: Herbivore) -> None:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            if other.health <= 0:
+                other.die()
```
DevSpaciX, line 31: If you add everytime self to alive list ( line 8 ) , you can be sure that your animal already in the `list of alive animals` is not you ? So you can call `remove` func here and delete `die()` method

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1112, opened 2024-05-27
'Solution'

Irina17191:


app/main.py
```diff
@@ -1 +1,32 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore: Animal) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health = herbivore.health - 50
+            if herbivore.health <= 0:
```
DevSpaciX, line 31: Large nesting of code is bad practice, you can just do a `return` and limit it to nesting at 1 level

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore: Animal) -> None:
```
DevSpaciX, line 28: Can this method become `static` ? Because you do not using self param : )

DevSpaciX (changes requested):

app/main.py
Irina17191: Yes, really)Thank you)

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Animal) -> None:
+        if not (isinstance(herbivore, Herbivore) and not herbivore.hidden):
+            return
```
Oleg-s-s, line 31: What if my animal is Herbivore and is not hidden. I think method should work in that case)

Oleg-s-s (changes requested):

DevSpaciX (approved):

PR #1117, opened 2024-05-28
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1 +1,37 @@
...
+        self._hidden = False
+        Animal.alive.append(self)
+
+    @property
+    def hidden(self) -> bool:
+        return self._hidden
+
+    def die(self) -> None:
+        if self in Animal.alive:
+            Animal.alive.remove(self)
```
DevSpaciX, line 16: You never gonna need it ?)) Remove this func if you do not using it

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self._hidden = False
+        Animal.alive.append(self)
+
+    @property
+    def hidden(self) -> bool:
+        return self._hidden
```
Oleg-s-s, line 12: I mean, its ok. But why you decided to make it as private attr?

Oleg-s-s (approved):

PR #1121, opened 2024-05-30
Solution

Katy4S:


app/main.py
```diff
@@ -1 +1,33 @@
...
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name: str = name
+        self.health: int = health
+        self.hidden: bool = hidden
+        self.__class__.alive.append(self)
+
+    def __str__(self) -> str:
+        return f"{{Name: {self.name}, " \
+            f"Health: {self.health}, Hidden: {self.hidden}}}"
```
vsmutok, line 16: better to use `()` then `\`

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if isinstance(target, Carnivore) or target.hidden:
+            return
+        target.health -= 50
+        if target.health <= 0:
+            self.__class__.alive.remove(target)
```
vsmutok, line 33:
```suggestion
            self.alive.remove(target)
```

vsmutok, line 33:
```suggestion
    def bite(self, target: Animal) -> None:
        if isinstance(target, Herbivore) and target.hidden is False:
            target.health -= 50
            if target.health <= 0:
                self.alive.remove(target)
```

vsmutok (changes requested):

app/main.py
Katy4S: Please provide an example of the linebreak with (). Thank you!

app/main.py
Katy4S: Why should it be done? Thank you!

app/main.py
```diff
@@ -1 +1,32 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
+                self.__class__.alive.remove(target)
```
EdAlekseiev, line 32: Please, don't use `__class__` in public function, because it is private attribute. You can rewrite it without class
```suggestion
                self..alive.remove(target)
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,32 @@
...
+
+class Animal:
+    alive: List["Animal"] = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name: str = name
+        self.health: int = health
+        self.hidden: bool = hidden
+        self.__class__.alive.append(self)
```
vasylhnatiuk, line 12: do the same here

app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+from typing import List
+
+
+class Animal:
+    alive: List["Animal"] = []
```
```suggestion
    alive: List[Animal] = []
```

vasylhnatiuk: use `from __future__ import annotations`

vasylhnatiuk (changes requested):

vsmutok (approved):
Great!

PR #1122, opened 2024-05-30
solution

AndreiVed:


app/main.py
```diff
@@ -1 +1,38 @@
...
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is False:
+            self.hidden = True
+        else:
+            self.hidden = False
```
vsmutok, line 31: better to use `not` in this case

app/main.py
```diff
@@ -1 +1,38 @@
...
+            health: int = 100,
+            hidden: bool = False) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        Animal.if_alive(self)
+
+    @classmethod
+    def if_alive(cls, beast: Animal) -> None:
+        cls.alive += [beast] if beast.health > 0 else []
```
vsmutok, line 18: this method redundant here

app/main.py
```diff
@@ -1 +1,38 @@
...
+            self.hidden = True
+        else:
+            self.hidden = False
+
+
+class Carnivore(Animal):
+    def bite(self, beast: Animal) -> None:
+        if isinstance(beast, Herbivore) and beast.hidden is False:
+            beast.health -= 50
+        Animal.alive.remove(beast) if beast.health <= 0 else []
```
vsmutok, line 38: why `else []`?

vsmutok (changes requested):

app/main.py
AndreiVed: hello! if i remove else [] - flake8 take me an error:
Screenshot 2024-05-31 at 08 36 57
  E999 SyntaxError: expected 'else' after 'if' expression

AndreiVed: Screenshot 2024-05-31 at 08 42 47

app/main.py
AndreiVed: ok, i fixed it. but i thought, what we need validation of beast's health before addition to Animal.alive list
Screenshot 2024-05-31 at 08 53 19

app/main.py
AndreiVed: ok, i fixed it. but could you explain me why its better way?

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is not False:
```
vasylhnatiuk, line 24: Vasyl meant to write in one line with not  - `self.hidden = not self.hidden`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1131, opened 2024-06-11
Implement solution

anton-bilous:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if not isinstance(other, Herbivore) or other.hidden:
+            return
```
MNDonut, line 31: let's use a direct condition: if it's carnivore - return because in the reqs it's mentioned that this method shouldn't work for Carnivore only but if we add another animal, you'll need to modify this method because for a new animal it won't work.
Also a direct condition is easier to read

MNDonut (changes requested):

MNDonut (approved):

PR #1136, opened 2024-06-14
Solution

VovkOL:


YukihiroSM (changes requested):
Fix flake 8, please. Else looks good to me for now

app/main.py
```diff
@@ -1 +1,36 @@
...
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def remove_dead(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
+
+    @classmethod
+    def __str__(cls) -> None:
+        return str([repr(animal) for animal in cls.alive])
```
vsmutok, line 23: This method should not be a class method, as it logically should be used to provide a string representation of an instance of the Animal class. In your code, __str__ is written as a class method, but it does not provide a correct string representation for an instance of the class.
```suggestion
    @classmethod
    def __str__(self) -> str:
        return repr(self)
```

vsmutok (changes requested):

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1140, opened 2024-06-17
Solution

oksanaklw:


app/main.py
```diff
@@ -1 +1,32 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100, hidden: bool = False)\
+            -> None:
```
EdAlekseiev, line 5: Use parentheses () instead of backslashes \ to separate lines.
```suggestion
    def __init__(
        self,
        name: str,
        health: int = 100,
        hidden: bool = False
    ) -> None:
```

EdAlekseiev (changes requested):

app/main.py
oksanaklw: Done

vsmutok (approved):
Great!

PR #1141, opened 2024-06-18
Solution

Alex414414:


app/main.py
```diff
@@ -1 +1,35 @@
...
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}," \
```
vasylhnatiuk, line 16: use () instead of \

app/main.py
```diff
@@ -1 +1,35 @@
...
+               f" Health: {self.health}," \
+               f" Hidden: {self.hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
```
vasylhnatiuk, line 23: instead of if-else use self.hidden = not self.hidden

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #1144, opened 2024-06-19
Solution

andreykolomiec:


EdAlekseiev (changes requested):
Fix tests, please

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: str) -> None:
```
Dimosphen1, line 37: Indicate that you expect to obtain a `Herbivore` class instance for `animal` in type annotation, not `str`

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        if self not in Animal.alive:
```
Dimosphen1, line 12: Always append new animals to the `alive` class attribute, it is unlikely that `__init__` would be called on an instance that was initialized earlier

app/main.py
```diff
@@ -1 +1,41 @@
...
+            f" Health: {self.health},"
+            f" Hidden: {self.hidden}}}"
+        )
+
+    def is_alive(self) -> int:
+        return self.health > 0
+
+    @staticmethod
+    def remove_dead_animals() -> None:
+        Animal.alive = [animal for animal in Animal.alive if animal.is_alive()]
```
Dimosphen1, line 27: Since these methods are not used anywhere, remove them

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1145, opened 2024-06-20
implemented Predator, Herbivore and Animal classes 

HalynaRohatska:
implemented Predator and Herbivore classes that inherit from the Animal class

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,37 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def die(self) -> None:
+        if self in Animal.alive:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return (f"{{"
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}"
+                f"}}")
+
+    def __str__(self) -> None:
+        print(Animal.alive.__repr__())
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(victim: Animal) -> None:
+        if isinstance(victim, Herbivore) and victim.hidden is False:
+            victim.health -= 50
+        if victim.health <= 0:
+            victim.die()


app/main.py
```diff
@@ -1 +1,37 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def die(self) -> None:
+        if self in Animal.alive:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
```
MNDonut, line 14: it's better to use str, repr is for developer's purposes

app/main.py
```diff
@@ -1 +1,37 @@
...
+        return (f"{{"
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}"
+                f"}}")
+
+    def __str__(self) -> None:
+        print(Animal.alive.__repr__())
```
MNDonut, line 22: why do you call repr method of list?(alive is a list)

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
```
MNDonut, line 32: why staticmethod? we don't have such a requirement

app/main.py
```diff
@@ -1 +1,37 @@
...
+    def bite(victim: Animal) -> None:
+        if isinstance(victim, Herbivore) and victim.hidden is False:
```
MNDonut, line 34: you can use not victim.hidden

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
LiudmylaKulinchenko, line 7: You should be albe to pass hidden attribute when creating an Animal instance. You should make it similar to health, just use a default value

app/main.py
```diff
@@ -1 +1,36 @@
...
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}"
+                f"}}")
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+
+class Herbivore(Animal):
+
```
LiudmylaKulinchenko, line 26: Remove unnecessary empty line to have same code style throughout the code

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1148, opened 2024-06-25
Solution

juliazelena:


app/main.py
```diff
@@ -1 +1,40 @@
...
+        )
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is True:
+            self.hidden = False
+        else:
+            self.hidden = True
```
Dimosphen1, line 29: Implement this method in one line as mentioned in the checklist using `not`

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Animal) -> None:
```
Dimosphen1, line 35: Indicate in the type annotation that `other` should be explicitly `Herbivore` class instance

app/main.py
```diff
@@ -1 +1,40 @@
...
+        if isinstance(other, Herbivore) and other.hidden is False:
```
Dimosphen1, line 36: You could simplify this if condition
```suggestion
        if isinstance(other, Herbivore) and not other.hidden:
```

Dimosphen1 (changes requested):

PR #1152, opened 2024-06-27
successfully implement OOP realisation of animals with hide and bite …

MasakDirt:
…methods

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
+            return
+
+        if not herbivore.hidden:
+            herbivore.health -= 50
+
+        if herbivore.health <= 0:
+            super().alive.remove(herbivore)


app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
MNDonut, line 5: either all parameters in 1 line or 1 parameter - 1 line

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
```
MNDonut, line 11: it's better to use str when you want to name the object, repr is for developer's goals

app/main.py
```diff
@@ -1 +1,33 @@
...
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
MNDonut, line 21: You can simplify it using **not**

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
+            return
```
MNDonut, line 27: The method does not work **if it is another сarnivore**, or the herbivore is currently hiding.

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+        if not herbivore.hidden:
+            herbivore.health -= 50
+
+        if herbivore.health <= 0:
+            super().alive.remove(herbivore)
```
MNDonut, line 33: For class attribute you can either use class name to get it or it's better to use herbivore.alive..... because there may be 50 classes and remember the class name may be difficult

MNDonut (changes requested):

app/main.py
MNDonut: My bad, a bit confusing if statement. Let's make it more explicit and check if it's Carnivore -> return

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
```
MNDonut, line 7: parameters should starts from the next line

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore) or herbivore.hidden:
+            return
```
MNDonut, line 26: left a comment in a discussion above

oliasymak24 (approved):
You changed everything correct)

PR #1156, opened 2024-07-02
Solution

x0wer:


app/main.py
```diff
@@ -1 +1,42 @@
...
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def take_damage(self, amount: int) -> Callable:
```
Dimosphen1, line 18: Change your return type annotation, you do not return `Callable` here and in the `die` and `hide` methods as well

app/main.py
```diff
@@ -1 +1,42 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> Callable:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> int:
```
Dimosphen1, line 40: You do not return `int` here, correct your return type annotation

Dimosphen1 (changes requested):

PR #1160, opened 2024-07-03

vldyu:


app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def die(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> Herbivore:
+        self.hidden = not self.hidden
+        return self
```
EdAlekseiev, line 32: Why do you return self here? Delete it, please
```suggestion
    def hide(self) ->None:
        self.hidden = not self.hidden
```

EdAlekseiev (changes requested):

vsmutok (approved):
Great!

PR #1164, opened 2024-07-09
Implemented solution

dimak20:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __repr__(self) -> str:
+        return (f"{{"
```
MNDonut, line 5: It's better to use __str__ when you want to give a name for an object. __repr__ is mostly used to see detailed name of object when you work from console(like memory cell, type, etc)

app/main.py
```diff
@@ -1 +1,36 @@
...
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}"
+                )
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False)\
```
MNDonut, line 14: it's better to move parameters on the next line(including the first one). And also we don't use \

app/main.py
```diff
@@ -1 +1,36 @@
...
+            Animal.alive.append(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
```
MNDonut, line 29: Why classmethod if in the example this method is called from class instance, not class itself?

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def bite(cls, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Carnivore):
+            return
+        if herbivore.hidden is False:
+            herbivore.health -= 50
```
MNDonut, line 34: just **not herbivore.hidden**

MNDonut (changes requested):

MNDonut (approved):

PR #1166, opened 2024-07-11
Solution

NazarSerdiuk1:


app/main.py
```diff
@@ -1 +1,33 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
```
pavlopro, line 18: Simplify this to one line` self.hidden = not self.hidden`

app/main.py
```diff
@@ -1 +1,33 @@
...
+            self.hidden = False
+        else:
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if not isinstance(herbivore, Herbivore):
```
pavlopro, line 26: Please combine those two `if`'s in one

pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+
+            if herbivore.health <= 0:
+                super().alive.remove(herbivore)
```
pavlopro, line 27: There is no need to call super, just use `Animal `instead

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1168, opened 2024-07-13
solution

Morphin20th:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        Animal.alive.append(self)
+
+    def check_and_remove(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
```
vsmutok, line 21: no need to create this variable here

app/main.py
```diff
@@ -1 +1,34 @@
...
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
+
+
+class Carnivore(Animal):
+    def bite(self, obj: Herbivore) -> None:
```
vsmutok, line 31: `obj` not good name of variable in this case

vsmutok (changes requested):

app/main.py
Morphin20th: very long line

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def check_and_remove(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        name = self.name
+        health = self.health
+        hidden = self.hidden
+        return f"{{Name: {name}, Health: {health}, Hidden: {hidden}}}"
```
vsmutok, line 22:
```suggestion
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, "
            f"Hidden: {self.hidden}}}"
       )
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1169, opened 2024-07-13
Solution

PavelTym:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def __str__(cls) -> str:
+        return str([repr(animal) for animal in cls.alive])
```
vsmutok, line 24: why you need this??
The code is incorrect because the __str__ method should be an instance method, not a class method.

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1170, opened 2024-07-14
Solution

OleksiukStepan:
Created and implemented class:

- **Animal**
- **Herbivore** with **hide** function
- **Carnivore** with **bite** staticmethod

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,34 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,34 @@
...
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
```
MNDonut, line 15: it's better to use str, repr is for developer's goals

app/main.py
```diff
@@ -1 +1,34 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
```
MNDonut, line 29: why staticmethod? is there a requirement for it?

MNDonut (changes requested):

MNDonut (approved):

PR #1173, opened 2024-07-19
Solution

DSahalatyi:


app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(prey: Herbivore) -> None:
```
LiudmylaKulinchenko, line 29: This method can take not only Herbivore, you should use parent class for annotations

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1177, opened 2024-07-25
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+
+        if isinstance(animal, Herbivore) and animal.hidden is False:
+            animal.health -= 50
+
```
MNDonut, line 31: no need to do these extra blank lines. we often do it before and after loops / if statements that are in the middle of the code, not at the beginning

app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(animal: Animal) -> None:
+
+        if isinstance(animal, Herbivore) and animal.hidden is False:
```
MNDonut, line 29: just `not animal.hidden` is enough

MNDonut (changes requested):

MNDonut (approved):

PR #1178, opened 2024-07-25
Solution

zazmarga:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        result = ""
```
vasylhnatiuk, line 14: why do you need this line?

vasylhnatiuk (changes requested):

app/main.py
zazmarga: You are right. This line is left over from an old idea, I didn't notice. Fixed

app/main.py
```diff
@@ -1 +1,33 @@
...
+                 ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        result = "{" + (f"Name: {self.name}, Health: {self.health}, "
+                        f"Hidden: {self.hidden}") + "}"
+        return result
```
vsmutok, line 16:
```suggestion
    def __repr__(self) -> str:
        return (f"{{Name: {self.name}, "
                f"Health: {self.health}, "
                f"Hidden: {self.hidden}}}")
```
better to use `f` string

vsmutok (changes requested):

app/main.py
zazmarga: done

vsmutok (approved):
Good Job!

PR #1180, opened 2024-07-26
implement py-herbivores-and-carnivores

ValeriiaKyr:


app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False) -> None:
```
MNDonut, line 9: Closing bracket should be always on a separate line

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def __str__(cls) -> str:
+        return str([repr(animal) for animal in cls.alive])
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
MNDonut, line 30: So in order to swap value of this attribute, you can use "not" operator and do it without if statements ;)

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
```
MNDonut, line 35: just `not herbivore.hidden` is enough

MNDonut (changes requested):

MNDonut (approved):

PR #1184, opened 2024-07-26

AndriiHamasa:


app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100,
+                 hidden: bool = False
+                 ) -> None:
```
MNDonut, line 11: read please this message and fix everywhere ;)
https://mate.academy/chats/244801?selectedMessageId=3129519

app/main.py
```diff
@@ -1 +1,43 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, victim: Herbivore | Carnivore) -> None:
```
LiudmylaKulinchenko, line 35: You can use parent class for annotations to cover both cases

app/main.py
```diff
@@ -1 +1,43 @@
...
+        if victim.hidden or isinstance(victim, Carnivore):
+            return
```
LiudmylaKulinchenko, line 37: Use opposite condition to avoid blank return

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks great! 🔥

PR #1188, opened 2024-07-27
Solution

mihavryliuk:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: "Herbivore") -> None:
+
+        if herbivore.hidden is False and isinstance(herbivore, Herbivore):
+            herbivore.health -= 50
```
MNDonut, line 37: We rarely use is True, is False because we can check the same thing using "if not ....". "is" is mostly used when we check for None type or reference objects. It isn't incorrect, but let's follow good practices

MNDonut (changes requested):

MNDonut (approved):

PR #1195, opened 2024-07-28
Solution

dpiuro:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def __init__(
+            self,
+            name: str,
+            health: int = 100
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        if self not in Animal.alive:
+            Animal.alive.append(self)
```
MNDonut, line 13: It's impossible that self is in Animal.alive, thus this condition is redundant. When you create a new object, how it can be in this list? Like this condition makes sense when you check if an animal of this type or with this name is in that list, but when you just use "in" it's impossible because python uses object identifier to check it

MNDonut (changes requested):

MNDonut (approved):

PR #1196, opened 2024-07-28
Implement py-herbivores-and-carnivores

Maksym-Turenko:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(pray: Herbivore) -> None:
+        if isinstance(pray, Herbivore) and pray.hidden is False:
```
MNDonut, line 36: let's use not pray.hidden. We rarely use is False, is True, this operator is mostly used for None. It isn't incorrect, but let's follow good practices

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{"
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}"
+                f"}}"
+                )
```
LiudmylaKulinchenko, line 21: Fix formatting here, please

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(pray: Herbivore) -> None:
```
LiudmylaKulinchenko, line 35: If this function only takes Herbivores, why would you need a check in the next line? Use parent class for annotations here to cover all cases

LiudmylaKulinchenko (changes requested):

PR #1199, opened 2024-07-29
Solution

olena4reunited:


app/main.py
```diff
@@ -1 +1,36 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health = other.health - 50
```
MNDonut, line 34: you can simplify it using -=

app/main.py
```diff
@@ -1 +1,36 @@
...
+            if other.health <= 0:
```
MNDonut, line 35: you don't need to have this condition because you have the same in not_alive method

MNDonut (changes requested):

MNDonut (approved):

PR #1202, opened 2024-07-29
Solution

Sergi0bbb:


app/main.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 33: If this function only takes Herbivores, why would you need a check in the next line? Use parent class for annotations here to cover all cases

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1210, opened 2024-08-02
add solution

georgiy-nezbritskiy:


app/main.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other_animal: Animal) -> None:
+        if not other_animal.hidden and not isinstance(other_animal, Carnivore):
```
LiudmylaKulinchenko, line 27: You should check if isinstance of Herbivore, because in the future Animal could have more inherited classes than only Carnivore and Herbivore

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🥥

PR #1216, opened 2024-08-09
Solution

Serhii-Chubur:


app/main.py
```diff
@@ -1 +1,38 @@
...
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
```
vasylhnatiuk, line 25: use self.hidden = not self.hidden instead of if-else

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1234, opened 2024-08-22
add solution for task

nikvitkaua:


app/main.py
```diff
@@ -1 +1,36 @@
...
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
```
vasylhnatiuk, line 24:
```suggestion
            self.hidden = not self.hidden
```

vasylhnatiuk: use it instead of if-else

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1240, opened 2024-08-31
Solution

PonomaryovNikita:


app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
+    def bite(cls, other_animal: Herbivore) -> None:
```
pavlopro, line 35: Instead of other_animal use just animal as variable name

app/main.py
```diff
@@ -1 +1,40 @@
...
+        if (other_animal.hidden is False
+                and isinstance(other_animal, Herbivore)):
+            other_animal.health -= 50
+        if other_animal.health <= 0:
```
pavlopro, line 39: Your check for the health of animal should be on nested into first `if` statement

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1245, opened 2024-09-05
'Solution'

Obusker:


app/main.py
```diff
@@ -1 +1,38 @@
...
+    def get_bit(self) -> None:
+        self.health -= 50
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden is False:
```
vasylhnatiuk, line 27: do it in one line, use  `self.hidden =  not self.hidden`

vasylhnatiuk (changes requested):

PR #1251, opened 2024-09-15
'Solution'

KKvladys:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+    alive = []
+    hidden = False
```
AnyoneClown, line 3: Initialize hidden inside of `init` method with default value

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #1256, opened 2024-09-20
animal

dmytrik:


app/main.py
```diff
@@ -1 +1,34 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
AnyoneClown, line 6: Each parametr should be on a new line
Example:
```
def __init__(
            self,
            name: str,
            health: int = 100,
            hidden: bool = False
) -> None:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if (isinstance(herbivore, Herbivore)
+                and not herbivore.hidden):
```
AnyoneClown, line 32: Bad identation, try to remake it in one line

AnyoneClown (changes requested):

PR #1257, opened 2024-09-22
Created Animal, Herbivore and Carnivore classes and their methods

Mxbely:


app/main.py
```diff
@@ -1 +1,34 @@
...
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        elif not self.hidden:
+            self.hidden = True
```
Atikiho, line 22: You can shorten if-elif condtion in on line by giving self.hidden it's opposite value using not operator: `self.hidden = not self.hidden`

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+class Carnivore(Animal):
+    def bite(self, obj: Animal) -> None:
+        if isinstance(obj, Herbivore) and not obj.hidden:
+            obj.health -= 50
+            if obj.health <= 0:
+                Animal.alive = [
+                    animal
+                    for animal in Animal.alive
+                    if obj.name != animal.name
```
Atikiho, line 33: You don't need to use list comprehension here, you can use lists method `remove`

Atikiho (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
-# write your code here
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = False
```
AnyoneClown, line 8: What if we want to create Animal instance and we want it to be `hidden`?

Define hidden in arguments of `__init__` method and give a default value to it

app/main.py
```diff
@@ -1 +1,27 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, obj: Animal) -> None:
+        if isinstance(obj, Herbivore) and not obj.hidden:
+            obj.health -= 50
+            if obj.health <= 0:
+                Animal.alive.remove(obj)
```
AnyoneClown, line 27: Let's give more descriptive name for variable. Use `animal`, or `victim` for example

AnyoneClown (approved):
Good job!

PR #1264, opened 2024-09-30
Solution

VladyslavBon:
I don't understand what's wrong, I don't know how to pass last test with print(Animal.alive)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,28 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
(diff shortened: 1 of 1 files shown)


skyfoxwork:
Hi! I think you should use  some like this:

    def __repr__(self):
        
for flake8: 

    def __repr__(self) -> str:
        return (
            f"{{Name: {self.name}, "
            f"Health: {self.health}, "
            f"Hidden: {self.hidden}}}"
        )

skyfoxwork (approved):
if you will use:
def __repr__(self) -> str:
    return (
        f"{{Name: {self.name}, "
        f"Health: {self.health}, "
        f"Hidden: {self.hidden}}}"
    )
    
    I think it will work !

Nattalli (approved):

PR #1265, opened 2024-10-01
Solution-py-herbivores-and-carnivores

skyfoxwork:


VladyslavBon (approved):
Looks good, well done)

app/main.py
```diff
@@ -1 +1,38 @@
...
+            f"Hidden: {self.hidden}}}"
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
Nattalli, line 28: better write it in one line:
`self.hidden = not self.hiden`

Nattalli (changes requested):

app/main.py
skyfoxwork: cool solution !!!

Nattalli (approved):

PR #1269, opened 2024-10-01
implement py-herbivores-and-carnivores

yuliiapyrih:


app/main.py
```diff
@@ -1 +1,43 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    @staticmethod
+    def print_alive() -> str:
+        return f'[{", ".join([repr(animal) for animal in Animal.alive])}]'
```
LiudmylaKulinchenko, line 25: No need to create a list inside join, it should work with a generator (just remove `[]`)

app/main.py
```diff
@@ -1 +1,43 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+    def die(self) -> None:
+        Animal.alive.remove(self)
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
LiudmylaKulinchenko, line 38: This function could receive all animals, not just Herbivore. If it would receive only herbivores, you won't need to make a type check in the next line

LiudmylaKulinchenko (changes requested):

app/main.py
yuliiapyrih: Done

LiudmylaKulinchenko (approved):
Good job! 🌸

PR #1278, opened 2024-10-04
my_solution

OlhaStadnik:


app/main.py
```diff
@@ -1 +1,35 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+    @classmethod
+    def print_alive(cls) -> str:
+        for animal in Animal.alive:
+            return repr([animal])
```
molodsh1y1, line 20: it's unnecessary

molodsh1y1 (changes requested):

molodsh1y1 (approved):
but it works

molodsh1y1 (approved):

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, herbivore: Herbivore) -> None:
```
Nattalli, line 26: that function input can be Herbivore, so type should be `Animal`
```suggestion
    def bite(self, animal: Animal) -> None:
```

Nattalli, line 26: better make it staticmethod or  classmethod

Nattalli (changes requested):

Nattalli (approved):

PR #1285, opened 2024-10-06
solution

VolodymyrSemchysyn:


app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __str__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+    def __repr__(self) -> str:
+        return self.__str__()
```
YaroslavBordovoy, line 19: You can use __repr__ method instead of __str__ immediately,

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore) and herbivore.hidden is False:
```
YaroslavBordovoy, line 29: You can use @staticmethod there. Also, I would write "animal: Animal" instead of "herbivore: Herbivore" because this method takes carnivores too.

YaroslavBordovoy (changes requested):

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(animal: Animal) -> None:
+        if isinstance(animal, Herbivore) and animal.hidden is False:
```
Nattalli, line 27:
```suggestion
        if isinstance(animal, Herbivore) and not animal.hidden:
```

app/main.py
```diff
@@ -1 +1,31 @@
...
+            animal.health -= 50
+            if animal.health <= 0:
+                Animal.alive.remove(animal)
+                print(f"{animal.name} is dead")
```
Nattalli, line 31: remove print

Nattalli (changes requested):

Nattalli (approved):

PR #1292, opened 2024-10-07
Solution

Bondzik-S:


YaYaYaroslav (approved):
all good

app/main.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
```
Nattalli, line 29: that function have an `Animal` input value
```suggestion
    def bite(animal: Animal) -> None:
```

app/main.py
```diff
@@ -1 +1,33 @@
...
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
```
Nattalli, line 30: that's why you make checking here

Nattalli (changes requested):

Nattalli (approved):

PR #1302, opened 2024-10-09
Solution

matweyes:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
+    def update_alive(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
```
LLkaia, line 24: Don't u think that it is overkill to iterate through all instances just to remove 1? I think it is better to remove it in `bite` function

LLkaia (changes requested):

app/main.py
matweyes: agree, fixed

LLkaia (approved):
Good job!

PR #1333, opened 2024-10-30
Solution

IraMladanovych02:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
+    def bite(cls, goal: Animal) -> None:
+        if isinstance(goal, Herbivore) and not goal.hidden:
+            goal.health -= 50
+            goal.remove_if_dead()
```
vsmutok: Why you need two `Carnivore` classes?
please remove one of them

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,35 @@
...
+            Animal.alive.remove(self)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @classmethod
```
Dimosphen1: There is no need to make this method as a classmethod since `cls` is not called. You could make it static in the current implementation

app/main.py
```diff
@@ -1 +1,35 @@
...
+    def bite(cls, goal: Animal) -> None:
```
Dimosphen1: It is better to explicitly indicate that goal should be `Herbivore` class instance

app/main.py
```diff
@@ -1 +1,35 @@
...
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False  # Set hidden as an instance variable
+        Animal.alive.append(self)
+
+    def __len__(self) -> int:
+        return len(Animal.alive)
```
Dimosphen1: There is no need for `__len__` here, remove it

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #1391, opened 2024-12-04
Animals life cycle

valerii-kashpur:
- added Animal parent class
- added Herbivore class
- added Carnivore class


julia4406 (approved):
Cool work!

PR #1397, opened 2024-12-06
'Solution'

Dogober:


taraskhom (approved):
Everything is all right but you can simplify bite method by using and statement

app/main.py
```diff
@@ -1 +1,45 @@
...
+        return f"{{{self}}}"
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+            return
+        self.hidden = False
```
L1nk3rrr: It can be a one-liner, please read a checklist!

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Union[Carnivore, Herbivore]) -> Union[str, None]:
```
L1nk3rrr: You can simply use str | None, without Union, it's not needed here

app/main.py
```diff
@@ -1 +1,45 @@
...
+        if isinstance(other, Carnivore):
+            return "Carnivore won't bit Carnivore"
+        if isinstance(other, Herbivore) and other.hidden:
+            return "Carnivore won't be able to bite hidden Herbivore"
+        if other.health > 0:
```
L1nk3rrr: Do you need this check? Every bite it is -50,it  looks like not required here

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def __str__(self) -> str:
+
+        return (f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}")
+
+    def __repr__(self) -> str:
+
+        return f"{{{self}}}"
```
L1nk3rrr: It looks like you need to use only __repr__

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Union[Carnivore, Herbivore]) -> Union[str, None]:
+        if isinstance(other, Carnivore):
+            return "Carnivore won't bit Carnivore"
+        if isinstance(other, Herbivore) and other.hidden:
+            return "Carnivore won't be able to bite hidden Herbivore"
```
L1nk3rrr: At task we don't have requirements to return some text.

> Carnivore has a bite method, which takes a herbivore object and decreases the object's health by 50. The method does not work if it is another сarnivore, or the herbivore is currently hiding.

L1nk3rrr (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+    @staticmethod
+    def bite(other: Union[Carnivore, Herbivore]) -> None:
+        if isinstance(other, Carnivore):
+            return
+        if isinstance(other, Herbivore) and other.hidden:
+            return
+        other.health -= 50
+        if other.health <= 0:
+            Animal.alive.pop(Animal.alive.index(other))
```
L1nk3rrr: at task you need to check only for is other instance of Herbivore and not hidden.
So you don't need return at all, simplify this method

L1nk3rrr: Also at 37 line ` Animal.alive.pop(Animal.alive.index(other))`

Simplify this line. We access class attributes through class name and it's better to use **remove when value is given** and **del when an index is given**. You do unnecessary work here

app/main.py
```diff
@@ -1 +1,37 @@
...
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Union[Carnivore, Herbivore]) -> None:
```
L1nk3rrr: at can be simlified to Animal, as all of them is child of Animal, don't needed Union

L1nk3rrr (changes requested):

app/main.py
Dogober: The method does not work if it is another сarnivore, or the herbivore is currently hiding. This is from condition of the task

Dogober: So I finally understand what exactly did you mean

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = False
+        self.health = health
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{"
+                f"Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
```
petrykivd: If you have already put these brackets in a separate line, do the same at the end, or vice versa.

petrykivd (changes requested):

petrykivd (approved):
Nice! 😎

PR #1401, opened 2024-12-08
AnimalBiteAndDie

maxakim0v:


app/main.py
```diff
@@ -1 +1,31 @@
...
+                f", Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
ruslanyemets: You can implement this method as static, because you are not using any self-attributes here

ruslanyemets (approved):

app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
L1nk3rrr: One attribute per line, when you have a lot of attributes. At checklist you have a good example of styling
```
    def long_function_name(
            var_one,
            var_two,
            var_three,
            var_four
    ) -> None:
```

app/main.py
L1nk3rrr: Ruslan is right, it must be a staticmethod

L1nk3rrr (changes requested):
Look at the comments, you are almost at the finish!

L1nk3rrr (approved):
Well done!

PR #1404, opened 2024-12-08
solved

Danylo-Ok:


kuinorymes (changes requested):
Hi, here are some flake8 issues, and I think You need to add last feature described in "readme.md". With __repr__ magic method, using interpolation(which will structure "Animal.alive" in suggested way) GL)

PR #1410, opened 2024-12-08
solution

MaksVakulenko:


app/main.py
```diff
@@ -1 +1,32 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
```
VovaGld, line 28: You could have used the staticmethod here because it doesn't use attributes from self

VovaGld (approved):

PR #1412, opened 2024-12-08
Solution

Professor-Douel:


MindDevastation (approved):

app/main.py
```diff
@@ -1 +1,43 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100
+                 ) -> None:
```
petrykivd: Fix code style, if you specify a parameter on a line with (, then the last parameter must have a ) !
2 options:
def __init__(self,
                 name: str,
                 health: int = 100) -> None:
 
 def __init__(
 self,
name: str,
 health: int = 100
) -> None:

petrykivd: Read please checklist (point 5) before pushing your solution.

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(other: "Animal") -> None:
```
petrykivd: If you do not have any problems importing this class, you do not need to use “Animal”, just Animal

app/main.py
```diff
@@ -1 +1,43 @@
...
+        if isinstance(other, Herbivore) and not other.hidden:
+            other.health -= 50
+            if other.health <= 0:
+                other.health = 0
+                Animal.remove_dead()
```
petrykivd: Using this method, you remove all animals with less than 0 health from the list at once, and I don't think this is right.
It would be more logical to remove a specific instance of an animal that was killed from the list, for example:
Animal.alive.remove(other)
or super().alive.remove(other)

app/main.py
petrykivd: other is not a good variable naming. In my opinion, better would be something like "target".

petrykivd (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Animal) -> None:
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
+                target.health = 0
```
L1nk3rrr: looks like we don't need this assign

app/main.py
```diff
@@ -1 +1,42 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self,
+                 name: str,
+                 health: int = 100) -> None:
```
L1nk3rrr:
```
One attribute per line, when you have a lot of attributes. At checklist you have a good example of styling
```
    def long_function_name(
            var_one,
            var_two,
            var_three,
            var_four
    ) -> None:
```

File: app/main.py [hunk 6]
```diff
@@ -1 +1,42 @@
...
+    def remove_dead(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
+
+    @classmethod
+    def get_alive_repr(cls) -> str:
+        return str([repr(animal) for animal in cls.alive])
+
+    @classmethod
+    def get_alive(cls) -> list:
+        return cls.alive
```
L1nk3rrr: these methods isn't needed due to task requirements

L1nk3rrr (changes requested):

petrykivd (approved):
GJ 👍

PR #1413, opened 2024-12-08
Solution

TretiakHelen:


toku666 (approved):
wp

app/main.py
```diff
@@ -1 +1,33 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str,
```
petrykivd, line 5: Please read checklist before pushing your solution.
Point 5, move the name to a new line :)

app/main.py
```diff
@@ -1 +1,33 @@
...
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if not herbivore.hidden:
+                herbivore.health -= 50
+                if herbivore.health <= 0:
+                    self.__class__.alive.remove(herbivore)
```
petrykivd, line 33: You can simplify checks because this level of nesting is not a good solution.
The first two checks can be combined into one, and the 3rd check may not be inside the previous one

petrykivd: Also you can use super().alive instead of self.__class__

petrykivd (changes requested):

PR #1414, opened 2024-12-09
Solution

MindDevastation:


app/main.py
```diff
@@ -1,10 +1,9 @@
...
 
-    def __init__(self,
-             name: str,
-             health: int = 100,
-             hidden: bool = False) -> None:
+    def __init__(self, name: str,
+                 health: int = 100,
+                 hidden: bool = False) -> None:
         self.name = name
         self.health = 100
```
bodiakof: try 'self.health = health' instead

app/main.py
```diff
@@ -1,10 +1,9 @@
...
         self.hidden = False
```
bodiakof: and 'self.hidden = hidden' here

bodiakof (changes requested):
Try changing what's in the comments

bodiakof (approved):
Well done!

PR #1415, opened 2024-12-09
letitsnow

toku666:


app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f" Health: {self.health}, Hidden: {self.hidden}}}")
```
Sichkaridze: In this line, you have small mistake. 
Extra space between f" and Health. You need to remove this. 
From this: 

f" Health: {self.health}, Hidden: {self.hidden}}}")

You must have this:

f"Health: {self.health}, Hidden: {self.hidden}}}")

And also i recommend to follow one codestyle, and change this code to look more like:
                f"Health: {self.health},"
                f"Hidden: {self.hidden}}}")

Sichkaridze (changes requested):

Sichkaridze (approved):

petrykivd (approved):
Nice 👍

Sichkaridze (approved):

PR #1421, opened 2024-12-09
Create task Herbivores and carnivores

ihorhalyskiy:


Arty-tech-dotcom (approved):

app/main.py
```diff
@@ -1 +1,41 @@
...
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        if self.health > 0:
+            Animal.alive.append(self)
+
+    def die(self) -> None:
+        if self.health <= 0:
+            if self in Animal.alive:
```
petrykivd, line 21: Make 1 of these 2 checks to reduce the level of nesting

app/main.py
```diff
@@ -1 +1,41 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+            if target.health <= 0:
+                target.die()
```
petrykivd, line 41: You check if health is less than 0 in 2 places, maybe you should remove this check from somewhere?

petrykivd (changes requested):

PR #1443, opened 2024-12-20
solved 2

Danylo-Ok:


ruslanyemets (approved):
Good job! But you can make it simplier, without getter and setter.

PR #1452, opened 2024-12-28
implement Animal, Herbivore and Carnivore classes

yarzhuravlov:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        self.check_is_alive()
+
+    def check_is_alive(self) -> None:
+        if self.health <= 0:
+            Animal.alive.remove(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                + f"Health: {self.health}, "
+                + f"Hidden: {self.hidden}}}")
```
oliasymak24, line 30: You can use triple quotes for such cases

oliasymak24 (approved):
Good job!

app/main.py
yarzhuravlov: Got it, thanks for the advice.

PR #1455, opened 2025-01-03
implement parent class "Animal" with two child classes "Herbivore" an…

ihetay:
…d "Carnivore"


ihetay (commented):
corrected Animal.remove_animal(herbivore) to herbivore.remove_animal()

PR #1473, opened 2025-01-15
Solution

Nikita-pyth:


app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
```
oliasymak24: be consistent, use default values in same way

app/main.py
```diff
@@ -1 +1,25 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
```
oliasymak24, line 16: In such cases you can use triple quotes, and pay attention to your formatting

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, prey: Herbivore) -> None:
```
oliasymak24: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,25 @@
...
+        if isinstance(prey, Herbivore) and not prey.hidden:
+            prey.health -= 50
+            if prey.health <= 0:
+                Animal.alive.remove(prey)
```
oliasymak24, line 30: Add a message that animal is dead

oliasymak24 (changes requested):

Y-Havryliv (approved):
gj!

PR #1480, opened 2025-01-18
Class Animal.

kram3ko:
Herbivore inherit  from Animal,
Carnivore from Animal. Completed logic of classes.


app/main.py
```diff
@@ -1 +1,38 @@
...
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+    health = 100
+    hidden = False
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
oliasymak24: Each parameter must be from new line

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+class Herbivore(Animal):
+
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> Herbivore:
```
oliasymak24: you don't need self here, so make this method static

app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+    health = 100
+    hidden = False
```
oliasymak24: This is renudant, move to _init_

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}"
+                f", Health: {self.health}, Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> bool:
+        self.hidden = not self.hidden
+        return self.hidden
```
oliasymak24: No need for return

oliasymak24 (changes requested):

app/main.py
kram3ko: used instead dataclass. But thanks understand.

app/main.py
kram3ko: set it as staticmethod.

app/main.py
kram3ko: removed duplicate code.. Understand mistake. Thanks

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+@dataclass
+class Animal:
+    alive = []  # class attribute
+
+    name: str
+    health: int = 100
+    hidden: bool = False
+
+    # add instance to Class list(alive) after init
```
oliasymak24, line 13: You need always clean up code and deleting comments

oliasymak24 (approved):

app/main.py
kram3ko: Ok. Thank you. understand.

PR #1523, opened 2025-02-17
Completed task herbivores-and-carnivores

Yurkiv313:


app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(obj: Herbivore) -> None:
+        if isinstance(obj, Herbivore) and not obj.hidden:
+            obj.health -= 50
+            obj.check_health()
```
Y-Havryliv: This function checks the animal's health, but it also removes the animal from the 'alive' list if its health reaches zero.  The function name suggests it only "checks" health, but it actually modifies the state of the class by removing the instance from the list.  This is a hidden side effect, making the function responsible for two tasks instead of one. Ideally, functions should do 1 task

Y-Havryliv (changes requested):

oliasymak24 (approved):

PR #1531, opened 2025-02-21
Solution

HappyDen08:


app/main.py
```diff
@@ -1 +1,32 @@
...
+        return str(cls.alive)
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, prey: "Animal") -> None:
```
PrimeGlorious: The bite() method does not use self, so you must make it a static to eliminate warnings in pytest.

PrimeGlorious (changes requested):

PR #1547, opened 2025-02-23
'py-herbivores-and-carnivores_v1'

OlenVe:


app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if not self.hidden:
+            self.hidden = True
+        else:
+            self.hidden = False
+
```
yarms1, line 23: use "self.hidden = not self.hidden" instead of "if" and "elif"

iSevenpwnz (approved):

app/main.py
OlenVe: Thank you. I'll rewrite.

PR #1566, opened 2025-02-27
Solution

BuryiOstap:


app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, grass_animal: Herbivore) -> None:
+        if not grass_animal.hidden and isinstance(grass_animal, Herbivore):
```
```suggestion
    def bite(self, grass_animal: Herbivore) -> None:
        if isinstance(grass_animal, Herbivore) and  not grass_animal.hidden:
```
The i`sinstance(grass_animal, Herbivore)` check should be performed before accessing `grass_animal.hidden `to ensure that `grass_animal` is indeed an instance of `Herbivore`.

vsmutok (changes requested):

app/main.py
BuryiOstap: thanks

vsmutok (approved):
Looks Good!

PR #1567, opened 2025-02-27
Solution

Nataliia0809:


vsmutok (changes requested):
Tests Failed!

app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, health: int = 100,
+                 hidden: bool = False) -> None:
```
Dimosphen1, line 5: It is better to place all place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

Dimosphen1 (approved):
Looks good to me, well done! One comment is left for consideration in the future

PR #1594, opened 2025-03-17
Solution

Sashazmei:


app/main.py
```diff
@@ -1 +1,64 @@
...
+        "'lion.name' should equal to 'Lion King'"
+    assert lion.health == 100, \
+        "'lion.health' should equal to 100'"
+    assert lion.hidden is False, "'lion.hidden' should equal to False'"
+
+    print("Current alive animals:", Animal.alive)
+
+    assert len(Animal.alive) == 1, \
+        "Constructor should add created animal to 'Animal.alive'"
+
```
viktoria-rybenchuk: why did you add this test here?

app/main.py
```diff
@@ -1 +1,64 @@
...
+
+if __name__ == "__main__":
+    Animal.reset_alive()
+    lion = Carnivore("Simba")
```
viktoria-rybenchuk: this code is redundant, remove it

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+    def remove_dead(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
+
+    @classmethod
+    def reset_alive(cls) -> None:
+        cls.alive.clear()
+
+    @classmethod
+    def __str__(cls) -> str:
+        return str(cls.alive)
```
vsmutok, line 24: The `__str__` method should be an instance method rather than a class method, so it can represent the string representation of a single instance, not the class itself. Remove the @classmethod decorator.

app/main.py
```diff
@@ -1 +1,37 @@
...
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+    @classmethod
+    def remove_dead(cls) -> None:
+        cls.alive = [animal for animal in cls.alive if animal.health > 0]
+
+    @classmethod
+    def reset_alive(cls) -> None:
+        cls.alive.clear()
```
vsmutok, line 20: These methods are unnecessary here.

vsmutok (changes requested):

app/main.py
Sashazmei: I delated this two @classmethod but if i run pytest i have 2 failed tests

Sashazmei: Or i need to delate function too?

PR #1605, opened 2025-03-23
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 

Recheck everything. You literally saying that I didn't do it. When I did it exactly the same as you said

PR #1611, opened 2025-03-25
SOLUTION

Yevhenii0536:


app/main.py
```diff
@@ -1 +1,39 @@
-# write your code here
+class Animal:
+    alive = list()
```
MNDonut, line 2: just alive = [] is enough

app/main.py
```diff
@@ -1 +1,39 @@
...
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, "
+            f"Health: {self.health}, "
+            f"Hidden: {self.hidden}}}"
+        )
+
+    @classmethod
```
MNDonut, line 22: we don't do it in this way. remove this decorator. __str__ method accepts self parameter

MNDonut (changes requested):

PR #1614, opened 2025-03-26
Solution

farzamgt:


app/main.py
```diff
@@ -1 +1,28 @@
...
+class Animal:
+    alive = []
+
+    def __init__(
+            self, name: str, health: int = 100, hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
```
Tolik-code: Better create a class method and use it there, like self.add_animal (add_animal cls method)

app/main.py
```diff
@@ -1 +1,28 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
Tolik-code: Since there is no usage of self, you can use there staticmethod.

app/main.py
```diff
@@ -1 +1,28 @@
...
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                Animal.alive.remove(herbivore)
```
Tolik-code: The logic of taking damage seems more general than only for Herbivore. It would be better to move it to the Animal class and create an Animal method like "take_damage".

Tolik-code (approved):

PR #1745, opened 2025-06-26
inheritance

SerhiiBasok:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+
+    alive = []
+
+    def __init__(self, name: str, health: int = 100) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
```
Orixara: I think it would be better to add attribute hidden in def __init__, like hidden: bool = False

app/main.py
```diff
@@ -1 +1,36 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"{{Name: {self.name}, Health: {self.health}, "
+            f"Hidden: {self.hidden}"
+            f"}}"
+        )
```
Orixara: it would have been cleaner if you had used something like that:
image

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def bite(self, herbivore: "Herbivore") -> None:
+        if not isinstance(herbivore, Herbivore):
+            return
+        if herbivore.hidden:
+            return
+        herbivore.health -= 50
+        if herbivore.health <= 0:
+            herbivore.health = 0
+            if herbivore in Animal.alive:
+                Animal.alive.remove(herbivore)
```
Orixara: You can make it shorter and without returns, the first and second “if” is irrelevant you can do one with "and" insteed. Also it can be @staticmethod because you didn't use self there and line with herbivore.health = 0 is irrelevant due to the fact that we removed animal at all

Orixara (changes requested):

PR #1749, opened 2025-06-26
Solution

MakovM:


app/main.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    def bite(self, target: str) -> None:
```
viannik: Better make sure the bite method explicitly requires an instance of Herbivore (the parameter is named target: str, but expects an object)

viannik (approved):
Overall, the code matches the README’s OOP structure and method behaviors, and implements all required features. The implementation is clean and readable.

PR #1752, opened 2025-06-26
Develop

Artcrafterrra:


app/main.py
```diff
@@ -1 +1,36 @@
-# write your code here
+class Animal:
+    alive = []
```
yagginov: Forgot about the type annotation here
`alive: list = []`
or
`alive: list[Animal] = []`

app/main.py
```diff
@@ -1 +1,36 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    @staticmethod
+    def bite(target: Animal) -> None:
+        if isinstance(target, Carnivore):
+            return  # dont bite herbivore
```
yagginov: Here, if we pass, for example, an instance of the Animal class, your code will work, because there is no check for such a case.

yagginov (approved):

PR #1763, opened 2025-06-27
created class Animal with subclasses

trellle:


app/main.py
```diff
@@ -1 +1,37 @@
...
+                + f"Hidden: {self.hidden}"
+                + "}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
MNDonut: it just swaps value when you call this method, so you can use here `not` operator

MNDonut (changes requested):

MNDonut (approved):

PR #1791, opened 2025-07-10
Solution

Evgeniya0208:


app/main.py
```diff
@@ -1 +1,31 @@
-# write your code here
+class Animal:
+    hidden = False
```
vsmutok: `hidden` should be an attribute of the object, not the class.

vsmutok (changes requested):

app/main.py
Evgeniya0208: fixed

vasylhnatiuk (approved):

PR #1827, opened 2025-08-11
donetask

lukianlukian:


app/main.py
```diff
@@ -1 +1,58 @@
-# write your code here
+from __future__ import annotations
+
+
+class AliveList(list):
+    def __repr__(
+        self,
+    ) -> str:
```
viktoria-rybenchuk: write it in one line for better readability, fix in all places

app/main.py
```diff
@@ -1 +1,58 @@
...
+        return "[" + ", ".join(
+            f"{{Name: {animal.name}, Health: {animal.health}, "
+            f"Hidden: {animal.hidden}}}"
+            for animal in self
+        ) + "]"
```
viktoria-rybenchuk: use f-string

viktoria-rybenchuk (changes requested):

PR #1842, opened 2025-08-24
Solution

kovalevv:
Adding solution to 'Animals' task


kovalevv:
That was one of the toughest tasks so far, where I had to find out proper solution.

PR #1878, opened 2025-09-22
Solution

delirium95:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.name = name
+        self.hidden = hidden
+        if self not in Animal.alive:
+            Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, Hidden: {self.hidden}}}")
+
+    def die(self) -> None:
```
vakt159, line 25: don`t think, that this method is necessary, you can just delete Animal from list in method Bite. But if AI ok with this, then me too

vakt159 (approved):

PR #1913, opened 2025-10-16
Solution

viloker:


app/main.py
```diff
@@ -1 +1,39 @@
...
+                 hidden: bool = False) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __sub__(self, number: int) -> int:
+        self.health -= number
+        if self.health <= 0:
+            print(1)
```
vsmutok: This method is redundant for this task.
Please remove it.

app/main.py
```diff
@@ -1 +1,39 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
+
+
+class Carnivore(Animal):
+    @classmethod
+    def bite(cls, animal: Animal | Herbivore | Carnivore) -> None:
+        if animal.hidden is False and not isinstance(animal, Carnivore):
+            animal -= 50
```
vsmutok: This method should not be a @classmethod

Also, follow the advice given to you by the AI mentor.
> This does not follow the description requirement: "Carnivore has a `bite` method, which takes a herbivore object and decreases the object's health by 50. The method does not work if it is another сarnivore, or the herbivore is currently hiding." Your `bite` is a `@classmethod` and checks `not isinstance(animal, Carnivore)` instead of explicitly checking `isinstance(animal, Herbivore)`. That allows non-Herbivore, non-Carnivore objects to be affected. Update `bite` to ensure it only acts when `isinstance(target, Herbivore)` and `not target.hidden`, and subtract exactly 50 health. Also prefer an instance method so the API matches the examples (e.g., `lion.bite(rabbit)`) from the description .

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
+
+
+class Carnivore(Animal):
+
+    def bite(self, animal: Animal | Herbivore | Carnivore) -> None:
```
```suggestion
    def bite(self, animal: Herbivore) -> None:
```

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = True if self.hidden is False else False
```
```suggestion
        self.hidden = not self.hidden
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1924, opened 2025-10-26
add solution

VadymVynokur:


app/main.py
```diff
@@ -1 +1,32 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, herbivore: Herbivore) -> None:
```
petrykivd: Is target for bite can be only herbivore? 😃

app/main.py
```diff
@@ -1 +1,32 @@
...
+        if isinstance(herbivore, Herbivore) and not herbivore.hidden:
+            herbivore.health -= 50
+            if herbivore.health <= 0:
+                Animal.alive.remove(herbivore)
```
petrykivd: You could call the parent class a little differently to remove the animal from the alive list

petrykivd (changes requested):

app/main.py
VadymVynokur: according to the task "Carnivore has a bite method, which takes a herbivore object and decreases the object's health by 50. The method does not work if it is another сarnivore, or the herbivore is currently hiding."

petrykivd: So it can be carnivore, but method does not work ;) 
Let's change type hint?

petrykivd (changes requested):

app/main.py
VadymVynokur: What's the point of giving a hint about the type that the method won't work with?

petrykivd (approved):
Well done! 😎

PR #1928, opened 2025-10-28
Solution

leemonch1k-space:


app/main.py
```diff
@@ -1 +1,37 @@
...
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if not herbivore.hidden:
+                herbivore.health -= 50
+            if herbivore.health <= 0:
+                Animal.alive.remove(herbivore)
```
VadymVynokur, line 38: here u can use super(). instead of Animal.

app/main.py
```diff
@@ -1 +1,37 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(herbivore: Herbivore) -> None:
+        if isinstance(herbivore, Herbivore):
+            if not herbivore.hidden:
```
VadymVynokur: imo, would be better to use only one if here:
if isinstance(herbivore, Herbivore) and not herbivore.hidden:

VadymVynokur (approved):

PR #1929, opened 2025-10-28
Added solution

vadym-zahrebelnyi:


app/main.py
```diff
@@ -1 +1,74 @@
...
+    """A class to represent a carnivore."""
+
+    @staticmethod
+    def bite(animal: Herbivore) -> None:
+        """Bite a herbivore.
+
+        :param animal: The herbivore to bite.
+        :type animal: Herbivore
+        """
+        if not (isinstance(animal, Carnivore) or animal.hidden):
```
stanislav-kishechko: I would advise simplifying this because it is a little difficult to read.

app/main.py
```diff
@@ -1 +1,74 @@
...
+            animal.health -= 50
+            if animal.health <= 0:
+                animal.__class__.alive.remove(animal)
```
stanislav-kishechko: Before deletion, you need to check whether the animal is on the list.

stanislav-kishechko (approved):

PR #1933, opened 2025-10-29
Completed task

ostapshpeha:


app/main.py
```diff
@@ -1 +1,29 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
+        if not animal.hidden and isinstance(animal, Herbivore):
+            animal.health -= 50
+            if animal.health <= 0:
+                Animal.alive.remove(animal)
```
VadymVynokur, line 34: don`t use Animal. ...
better use super(). ...

app/main.py
```diff
@@ -1 +1,29 @@
...
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
```
VadymVynokur: at QnN we decided to use animal: Animal

app/main.py
```diff
@@ -1 +1,29 @@
...
+        if not animal.hidden and isinstance(animal, Herbivore):
```
VadymVynokur, line 31: change the order of the checks, otherwise there may be an error if you pass Carnivore.
Remember that the checks are done from left to right

VadymVynokur: nvm, i forgot that all Animal have .hidden by deffolt

VadymVynokur (approved):

app/main.py
ostapshpeha: Thank you for this review!

PR #1942, opened 2025-10-30
Solution

YanTumar:


app/main.py
```diff
@@ -1 +1,34 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if not isinstance(other, Herbivore):
+            return
+        if other.hidden:
+            return
```
Taras-Ivashchuk, line 27: Yan, please re-write 1 check instead of 2 checks  using logical operator

Taras-Ivashchuk (changes requested):

app/main.py
YanTumar: Fix it in solution 2

PR #1947, opened 2025-10-31
Develop

vadimchubok:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        )
+
+
+class Herbivore(Animal):
+
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
```
ostapshpeha: def hide(self) -> None:
        self.hidden = not self.hidden
        
  there is no need in "if"

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}"
+                )
+
+    def __str__(self) -> str:
```
ostapshpeha: no need in this magic method, u already have __repr__

ostapshpeha (approved):

PR #1948, opened 2025-10-31
Herbivores and carnivores

n-korobko:


tigersaymon (changes requested):
great job, not much to change, lines 38-41 can be combined, and recheck line 8, there is a better way to split a long string

PR #1955, opened 2025-11-01
created class animal, carnivore, herbivore

AlexMishKo:


dmytro-malyk-dm (approved):
Good work.
Pay attention to the annotation type when creating a class attribute, you should check the checklist.

PR #1981, opened 2025-11-21
Update main.py

Sashaaaaaa6:


app/main.py
```diff
@@ -1 +1,36 @@
...
+        self.health = health
+        self.hidden = False
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (
+            "{Name: " + self.name
+            + ", Health: " + str(self.health)
+            + ", Hidden: " + str(self.hidden) + "}"
+        )
```
Dimosphen1, line 15: Consider using a multi-line f-string for better readability rather than string concatenation

app/main.py
```diff
@@ -1 +1,36 @@
...
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, other: Animal) -> None:
+        if isinstance(other, Carnivore):
+            return
+        if other.hidden:
+            return
```
Dimosphen1, line 31: You can unite both if conditions into one and avoid using return here. For that, check that the other is an instance of `Herbivore` and is not hidden. Only in that case, perform the logic with decreasing the health.

app/main.py
```diff
@@ -1 +1,36 @@
...
+        other.health -= 50
+        if other.health <= 0:
+            Animal.alive = [
+                a for a in Animal.alive if a is not other
```
Dimosphen1, line 35: Avoid using single-letter variable names in Python for better readability.

Dimosphen1 (changes requested):

PR #1983, opened 2025-11-23
Solution

Botvynko95:


app/main.py
```diff
@@ -1 +1,61 @@
...
+
+    @health.setter
+    def health(
+        self: "Animal",
+        value: int,
+    ) -> None:
+        self._health = max(0, value)
+
+        if self._health == 0 and self in Animal.alive:
+            Animal.alive.remove(self)
```
vsmutok: Why do you need this here?
In this task, the property and setter are unnecessary.

app/main.py
```diff
@@ -1 +1,61 @@
...
+
+
+class Animal:
+    alive: list["Animal"] = []
+
+    def __init__(
+        self: "Animal",
+        name: str,
+        health: int = 100,
+    ) -> None:
```
vsmutok, line 11:
```suggestion
class Animal:
    alive: list[Animal] = []

    def __init__(
        self:
        name: str,
        health: int = 100,
    ) -> None:
```

You don’t need to specify a type annotation for `self`.

vsmutok (changes requested):

Botvynko95:
I removed the property and setter, and I removed the annotation for self (I added it earlier because the AI wouldn't let the code pass the tests).

vsmutok (approved):
Great!

PR #1991, opened 2025-11-28
solved

meh-666:


meh-666:
@mateacademy-ai-mentor  I didn’t make the bite method static because it’s intended to be used by a specific carnivore instance, even though "self" isn’t used inside it. That’s why I added a comment to suppress the IDE warning suggesting that the method should be static.

app/main.py
```diff
@@ -1 +1,104 @@
...
+class Carnivore(Animal):
+    """
+    Carnivore class representing predators.
+    Carnivores can attack herbivores, reducing their health by 50.
+    Hidden herbivores and other carnivores cannot be attacked.
+
+    """
+
+    # noinspection PyMethodMayBeStatic
+    def bite(self, target: Union[Herbivore, Carnivore]) -> None:
```
```suggestion
    def bite(self, target: Herbivore) -> None:
```

app/main.py
```diff
@@ -1 +1,104 @@
...
+        :param target: Herbivore or Carnivore instance to attack
+        :return: None
+        """
+        if isinstance(target, Carnivore) or target.hidden:
+            return
+
+        target.health -= 50
+
+        if not target._is_alive():
+            target._die()
```
```suggestion
        if isinstance(target, Herbivore) and not target.hidden:
            target.health -= 50
            if not target._is_alive():
                target._die()
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2024, opened 2025-12-21
Solution

tarasmosiichuk01-ship-it:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+
+    @staticmethod
+    def bite(other: Animal) -> None:
```
```suggestion
    def bite(other: Herbivore) -> None:
```

vsmutok (changes requested):

app/main.py
tarasmosiichuk01-ship-it: Firstly I specified the type Herbivore, but ai-mentor in a previous check noted that the requirement implies the method should be able to handle being passed other Animal types (like a Carnivore) and not act on them. The isinstance(other, Herbivore) check on the next line is redundant if the input is always a Herbivore. Why is that?

vsmutok (approved):
Great!

PR #2064, opened 2026-01-13
'Solution'

Nahorni:


app/anilmal/herbivore.py
```diff
@@ -0,0 +1,11 @@
...
+from app.anilmal.animal_base import Animal
+
+
+class Herbivore(Animal):
+    def hide(self) -> Animal:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+        return self
```
vsmutok:
```suggestion
class Herbivore(Animal):
    def hide(self) -> None:
        self.hidden = not self.hidden
```

app/main.py
```diff
@@ -1 +1,28 @@
...
+rabbit.hide()
+lion.bite(rabbit)
+print(rabbit.health)
+
+print(rabbit in Animal.alive)
+
+pantera = Carnivore("Bagira")
+snake = Carnivore("Kaa")
+print(Animal.alive)
+Animal.alive.clear()
```
vsmutok: Please remove your manual testing from this module.

vsmutok (changes requested):
Also this task requires that all code be written in a single module, `main.py`.
Splitting code into multiple modules is a good practice and generally recommended, but in this specific case it is unnecessary because the task does not call for it.

vsmutok (approved):
Great!

PR #2185, opened 2026-03-02
Solution

Reverad:


app/main.py
```diff
@@ -1 +1,37 @@
-# write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
```
SerhiiHryniuk: Add type annotation here because AI will say that you do not have type annotation, because was having the same way, but AI returned

SerhiiHryniuk (changes requested):

SerhiiHryniuk (approved):

PR #2225, opened 2026-03-24
solved py-herbivores-and-carnivores

pawunder1992:


app/main.py
```diff
@@ -1 +1,34 @@
...
+                 hidden: bool = False) -> None:
+        self.health = health
+        self.name = name
+        self.hidden = hidden
+        self.alive.append(self)
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
```
lubkoserhii, line 15: You're trying to split the f-string into multiple lines by using f“” on each one, but they aren't joined by a + operator or a comma. This will cause a SyntaxError. Use a single f-string

app/main.py
```diff
@@ -1 +1,34 @@
...
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden is True :
+            self.hidden = False
+        else :
+            self.hidden = True
+
+
+class Carnivore(Animal):
+    @staticmethod
```
lubkoserhii, line 24: The bite method should not be static

lubkoserhii (changes requested):
Fix these errors and submit a new pull request. Pay special attention to the f-line (line 13-15).

app/main.py
pawunder1992: In Python, there is a rule of implicit string literal concatenation. If two strings (regular or f-strings) are placed next to each other, separated only by whitespace or a line break (within parentheses), Python automatically joins them into a single long string at compile time. Therefore, no SyntaxError will occur. Furthermore, if I add the + operator, it will violate point 3 of the checklist.md.

app/main.py
pawunder1992: done

pawunder1992: AI does not accept the changes and refers to the first point of checklist.md.

lubkoserhii (approved):

PR #2227, opened 2026-03-25
my solution py-herbivores-and-carnivores

zubak-vasil:


app/main.py
```diff
@@ -1 +1,38 @@
...
+            self,
+            name: str,
+            health: int = 100,
+            hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+
+        Animal.alive.append(self)
```
PavelPereverzev1, line 14: This way of adding an instance allows adding multiple duplicates of the same object. I think it's better to do this through a class method with uniqueness control.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+
+        if isinstance(target, Herbivore) and not target.hidden:
+            target.health -= 50
+
+            if target.health <= 0:
+                target.health = 0
+                if target in Animal.alive:
```
PavelPereverzev1, line 37: I don't think an instance should have the authority to directly modify a class attribute like Animal.alive. This logic should be encapsulated within class methods to avoid code duplication. As we add more classes, repeating this manual removal everywhere will become a maintenance problem.

A better approach would be to simply update the health attribute and let a dedicated internal mechanism (like a setter or a class method) handle the registry updates. This centralizes the 'death' logic in one place and ensures consistency across all animal types.

PavelPereverzev1 (approved):

PR #2246, opened 2026-04-03
Solution

dimaaa-cmd:


alisahoys (approved):
Clean inheritance, responsibilities are well separated

PR #2247, opened 2026-04-03
Inheritance solution

alisahoys:


app/main.py
```diff
@@ -1 +1,29 @@
-# write your code here
+class Animal:
+    alive = []
```
andriihrechko: I recommend using type hints here.

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+    def __repr__(self) -> str:
+        return (f"{{Name: {self.name}, "
+                f"Health: {self.health}, "
+                f"Hidden: {self.hidden}}}")
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
```
andriihrechko, line 21: The Animal class should store the attribute for all alive animals at the class level, not on self.

app/main.py
```diff
@@ -1 +1,29 @@
...
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Herbivore) -> None:
+        if animal.hidden is not True and isinstance(animal, Herbivore):
+            animal.health -= 50
+            if animal.health <= 0:
+                self.alive.remove(animal)
```
andriihrechko: You need to remove the value from the Animal class attribute, not from self.

andriihrechko (approved):

PR #2267, opened 2026-04-07
Solution

WachaX52:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        hidden: bool = False
+    ) -> None:
+        self.name = name
+        self.health = health
+        self.hidden = hidden
+        Animal.alive.append(self)
+
+    def __repr__(self) -> str:
+        return f"{{Name: {self.name}, Health: {self.health}, " \
+               f"Hidden: {self.hidden}}}"
```
petrykivd: Avoid using \ for line continuation — use parentheses instead
Braces are explicit, reliable, and are the standard according to PEP 8.

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, target: Animal) -> None:
+        if isinstance(target, Herbivore) and not target.hidden:
```
petrykivd: Let's also check if the herbivore is alive before biting it?

petrykivd (changes requested):

AnyoneClown (approved):
LGTM

PR #2302, opened 2026-05-02
solution

ulamelnyk:


app/main.py
```diff
@@ -1 +1,40 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        self.hidden = not self.hidden
+
+
+class Carnivore(Animal):
+    def bite(self, animal: Animal) -> None:
```
```suggestion
    def bite(self, animal: Herbivore) -> None:
```

The bite method's parameter annotation should be Herbivore type instead of Animal, as the task specifies this method takes a herbivore object.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2359, opened 2026-06-22
'My_Solution'

Oleksandr-131:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    @property
+    def health(self) -> int:
+        return self._health
+
+    @health.setter
+    def health(self, value: int) -> None:
+        self._health = value
+        if self._health <= 0:
+            Animal.remove_dead(self)
```
viktoria-rybenchuk, line 21: why do you use 2 method with the same name?

viktoria-rybenchuk (changes requested):

vsmutok (approved):
Great!!

PR #2396, opened 2026-07-15
Added herbivores and carnivores task solution

annsummmer:


app/main.py
```diff
@@ -1 +1,53 @@
...
+        self.health -= 50
+
+    @staticmethod
+    def handle_death(target: Animal) -> None:
+        """Removes instance from the list."""
+        Animal.alive = [
+            animal
+            for animal in Animal.alive
+            if animal is not target
+        ]
```
hladiivolodymyr: Using a list comprehension here seems unnecessary since you only need to remove a single item from the list. In this case, using list.remove() is a simpler and more readable solution.

tests/test_main.py
```diff
@@ -136,6 +136,7 @@ def test_print_animal_alive():
     output = "[{Name: King Lion, Health: 100, Hidden: False}, " \
              "{Name: Bagira, Health: 100, Hidden: False}, " \
              "{Name: Susan, Health: 100, Hidden: False}]\n"
+    print(out)
```
hladiivolodymyr: Please remove this print statement

hladiivolodymyr (changes requested):

tests/test_main.py
annsummmer: Fixed 1be0d36237c04ce073653d28afdfc350c4b2156a

app/main.py
annsummmer: Fixed 1be0d36237c04ce073653d28afdfc350c4b2156a

Sn1kls (approved):
Good job! Approve

PR #2423, opened 2026-08-07
Implemented classes

sarychev90:


Sn1kls (approved):
Approve – one nit: biting an already-dead herbivore crashes (`Animal.alive.remove()` on a missing item). Guard it: `if herbivore.health <= 0 and herbivore in Animal.alive: Animal.alive.remove(herbivore)`

PR #2445, opened 2026-08-13
do-the-task

ArtemSl0:


MaksimDonchyk (approved):
It`s ok but don`t over compicated by adding 
@property
    def health(self) -> int:
        return self._health
@health.setter
    def health(self, value: int) -> None:
        self._health = value

PR #2460, opened 2026-08-14
Solution

AndrewBudkin1:


app/main.py
```diff
@@ -1 +1,34 @@
...
+        )
+
+
+class Herbivore(Animal):
+    def hide(self) -> None:
+        if self.hidden:
+            self.hidden = False
+        else:
+            self.hidden = True
+
```
OkeNNV, line 24: It is better to `self.hidden = not self.hidden` instead of if statement. It does the same thing, but code is much shorter.

OkeNNV (approved):
Everything is correct, but you can simplify the if statement using `self.hidden = not self.hidden.`

